PR #28, opened 2021-01-08
Add producer support

themis-pyrgiotis:


stavshamir (changes requested):
Hi Themis, this looks in the right direction! I would really want however to have a `ProducerChannelsScanner`. Then we would not have to change anything in the `DefaultAsyncApiService`.

We can achieve this if you create a class that looks roughly like this:
```java
@Component
public class ProducerChannelsScanner implement ChannelsScanner {

    private final AsyncApiDocket docket;
    private SchemasService schemasService;

    @Override
    public Map<String, Channel> scan() { ... }

}
```

This will be picked up automatically in the `DefaultChannelsService` and handled there.
Once you make this change (and maybe add a test if you have the time) then we can merge this feature.

stavshamir:
Thanks for the contribution Themis!

themis-pyrgiotis:
You’re welcome Stav!

Issue #31, opened 2021-10-09
publish/subscribe semantics are mixed up


stavshamir:
# Description
Currently, listeners are mapped as `subscribe` channels, and producers are mapped as `publish` channels. Apparently, it should be the other way around.
See https://www.asyncapi.com/blog/publish-subscribe-semantics.

Thanks to https://github.com/kalarani for notifying me.

# Fix
1. The `subscribe` operation in [AbstractChannelScanner::buildChannel](https://github.com/springwolf/springwolf-core/blob/master/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/AbstractChannelScanner.java) should be replace with `publish`. 
2. Similarly, `publish` should be switched to `subscribe` in the [ProducerChannelScanner](https://github.com/springwolf/springwolf-core/blob/master/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/ProducerChannelScanner.java)
3. Recommended - add a unit test for `AbstractChannelScanner` and `ProducerChannelScanner`. For `AbstractChannelScanner`, this would require creating an inner class in the unit test the implements the abstract methods as simple as possible.
4. After the fix in the core project, I will release a new version, and then the tests in kafka plugin project should be fixed. Also, the UI project should be fixed accordingly.

kalarani:
Hi @stavshamir I'm working on this. Will raise a PR shortly.

PR #32, opened 2021-10-10
Fix for #31 - Publish/subscribe semantics are mixed up

kalarani:


stavshamir:
Thanks again for submitting the issue and for providing a fix!

Issue #38, opened 2022-01-21
IndexOutOfBoundsException when using bindings instead of queues


Cimballi:
There is an IndexOutOfBoundsException when using RabbitListener bindings instead of queues.

Here is the stack trace :
```
Caused by: java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:64)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:70)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:248)
	at java.base/java.util.Objects.checkIndex(Objects.java:372)
	at java.base/java.util.ArrayList.get(ArrayList.java:459)
	at io.github.stavshamir.springwolf.asyncapi.scanners.channels.RabbitChannelsScanner.getChannelName(RabbitChannelsScanner.java:45)
	at io.github.stavshamir.springwolf.asyncapi.scanners.channels.RabbitChannelsScanner.getChannelName(RabbitChannelsScanner.java:21)
	at io.github.stavshamir.springwolf.asyncapi.scanners.channels.AbstractChannelScanner.mapMethodToChannel(AbstractChannelScanner.java:81)
```

stavshamir:
Thanks for opening this issue and providing the stack trace!
I am not sure what you mean by using bindings instead of queues - do you mind adding an example of one of your `RabbitListener` annotated methods?

Cimballi:
Something like that :
```
    @RabbitListener(bindings = {
        @QueueBinding(
            exchange = @Exchange(name = "...", type = "topic"),
            key = "...",
            value = @Queue(name = "..."))
    })
```

stavshamir:
That's very helpful, thanks. I will try to release a patch in the next few days (if you'd like to provide a fix yourself, that's more than welcome, I'd be happy to help).

stavshamir:
Thanks for the fix @Omerbea! This will be included in the next release.

stavshamir:
Released in `springwolf-amqp-0.2.0`.

PR #44, opened 2022-03-04
IndexOutOfBoundsException when using bindings instead of queues #38

Omerbea:
according to the documentation bindings and queues are mutually exclusive so the list will contain the one that was decalred.


Omerbea:
@stavshamir

stavshamir:
Great contribution @Omerbea , thanks!

PR #49, opened 2022-03-19
Update README.md

ZachHubbs:
Was running this locally and found the documentation was outdated. This works with the most recent version of async api.


stavshamir:
Thanks for catching this!

PR #63, opened 2022-05-06
Ability of specifying base package for configuration beans

Stmated:
Ability of specifying base package for configuration beans, so components that are registered inside `@Configuration` can be found. This is because some projects simply do not use the `@Component` annotation, instead manually wire the beans.

In a way I think this overcomplicates things, and it would almost be easier to give ANT path pattern for which classes to search, and the code would just search all the methods of all the classes that match that pattern, with no regard to any class-level annotations at all.

Probable issues with this PR:
* It removes `@Configuration` from being matched in the regular base package scanning. So if a user has a `@KafkaListener` inside a `@Configuration` class, it would not be found.


Stmated:
It would probably also be better to change the docket to take a `ComponentScanner` instance instead of specifying the `basePackage`, so it is decoupled and you can give a custom `ComponentScanner` that works completely different depending on need. Like not wanting any scanning at all, and just giving your classes manually.

stavshamir:
Hi @Stmated , thanks for the PR! This does sound like a relevant use case that should be supported. 
I will look into it soon, in the meanwhile please merge from master as I just fixed a failing test.

springwolf-core/src/main/java/io/github/stavshamir/springwolf/configuration/AsyncApiDocket.java
```diff
@@ -18,9 +18,13 @@ public class AsyncApiDocket {
     /**
      * The base package to scan for listeners.
      */
-    @NonNull
     private String basePackage;
```
stavshamir, line 21: I think we should make the comments more clear, as `basePackage` and `configurationBasePackage` are very similar and may be confusing. 

I think this should change to: 
```java
/**
 * The base package to scan for listeners which are declared inside a class annotated with @Component or @Service.
*/
```

springwolf-core/src/main/java/io/github/stavshamir/springwolf/configuration/AsyncApiDocket.java
```diff
@@ -18,9 +18,13 @@ public class AsyncApiDocket {
...
 
+    /**
+     * The base package to scan for configurations that contain beans that contain listeners.
+     */
+    private String configurationBasePackage;
```
stavshamir, line 26: Like they comment above, I think this comment should be more specific:
```java
/**
* The base package to scan for @Configuration classes that contain beans that contain listeners. 
* Set this property if your listeners are declared in a class that is not annotated with @Component or @Service,
* and are declared in a class which is instantiated in a @Bean method in a @Configuration class.
*/
```

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/components/ComponentsScanner.java
```diff
@@ -9,13 +9,16 @@ public interface ComponentsScanner {
      * @param basePackage The root package to scan.
      * @return A set of found classes.
      */
-    Set<Class<?>> scanForComponents(Package basePackage);
+    default Set<Class<?>> scanForComponents(String basePackage) {
```
stavshamir, line 12: Let's remove the argument completely form `scanForComponents` and leave it without arguments. Then we can also remove the `AsyncAciDocket` wirings from `AbstractChannelScanner` etc. and have `DefaultComponentScanner` ue the docket directly.

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/components/DefaultComponentsScanner.java
```diff
@@ -17,28 +24,79 @@
 public class DefaultComponentsScanner implements ComponentsScanner {
```
stavshamir, line 25: Wire the `AsyncApiDocket` here instead of in `AbstrctChannelScanner`, `AbstractMethodLevelListener` and `ClassLevelKafkaListenerScanner`.

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/components/DefaultComponentsScanner.java
```diff
@@ -17,28 +24,79 @@
...
     }
 
     @Override
-    public Set<Class<?>> scanForComponents(String basePackage) {
-        log.debug("Scanning for component classes in {}", basePackage);
+    public Set<Class<?>> scanForComponents(String basePackage, String configurationBasePackage) {
+
+        Set<Class<?>> components;
+        if (StringUtils.isNotBlank(basePackage) && StringUtils.isNotBlank(configurationBasePackage)) {
+            log.debug("Scanning for component classes in configuration package {}, filtered by beans in {}", configurationBasePackage, basePackage);
```
stavshamir, line 36: I think the filtering part is going to be too confusing - if the user want to scan based on beans in a configuration class, let it scan all the beans. Let's remove the filtering logic.

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/components/DefaultComponentsScanner.java
```diff
@@ -17,28 +24,79 @@
...
+            throw new IllegalArgumentException("There must be either a basePackage or configurationBasePackage given");
+        }
 
+        return components;
+    }
+
+    private Set<Class<?>> scanForAutoConfigurationComponents(String basePackage) {
         ClassPathScanningCandidateComponentProvider provider = new ClassPathScanningCandidateComponentProvider(false);
         provider.addIncludeFilter(new AnnotationTypeFilter(Component.class));
+        provider.addExcludeFilter(new AnnotationTypeFilter(Configuration.class));
```
stavshamir, line 58: Why should we exclude `@Configuration`?

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/components/DefaultComponentsScanner.java
```diff
@@ -17,28 +24,79 @@
...
+            .filter(Optional::isPresent)
+            .map(Optional::get)
+            .map(configurationClass -> {
+
+                Set<Class<?>> beans = new LinkedHashSet<>();
+                for (final Method method : configurationClass.getDeclaredMethods()) {
+                    final Bean bean = AnnotationUtils.getAnnotation(method, Bean.class);
+                    if (bean != null) {
+                        final Class<?> returnType = method.getReturnType();
+                        final String beanPackage = returnType.getPackage().getName();
```
stavshamir, line 84: As mentioned above, I think this would be too confusing, let's remove the package filtering logic here.

stavshamir (changes requested):
@Stmated, that's a great contribution, well thought out and useful. I also appreciate providing new unit tests :) I have a few comments that I would like you to check and after the required changes I would gladly merge this!

stavshamir:
@Stmated, that's a great contribution, well thought out and useful. I also appreciate providing new unit tests :) I have a few comments that I would like you to check and after the required changes I would gladly merge this!

pontus-eliason:
I did the changes noted in this PR, but also did a few extra things (which you might disagree with, who knows):

* Changed so `AsyncApiDocket` takes a `ComponentsScanner` directly, but backwards compatible to previous basePackage way.
* Added a `ApplicationContextComponentsScanner` which searches the bean registry. What's good about it: Can find beans given by external libraries and by a `FactoryBean`.
* Separated the different `ComponentsScanners` into what they do, and created a default `DefaultClassPathComponentsScanner` that searches for `@Component` and `@Configuration` + `@Bean` annotations.

I did not make any changes to documentation, since it all works like it did previously, just with an option to swap the scanner if want to.

Stmated:
And that ^ was me. Having multiple accounts is a hassle.

stavshamir:
Sorry for the long time it took me to review! 
Wow, that's a **very** impressive contribution. Thanks a lot. I will merge this now (the build is failing on some java doc stuff, I suspect it's related to lombok - I won't bother you with fixing that, I can do it myself).

Since this is a big amount of content added to the library, that would be great if we could also document it, so if you are looking for further contributions, I'm opening an issue in the project documentation site for this additions.

You are also welcome to the discord server that's starting to attract some action, and please add me on linkedin (https://www.linkedin.com/in/stavshamir/).

Cheers!

PR #68, opened 2022-05-27
Fixed documentation

ogaudefroy:
👋🏻 a very minor contrib but still, I updated the documentation to be in sync with the current folder hierarchy.


stavshamir:
Small, but important :) Thanks for the fix!

PR #73, opened 2022-06-10
Chore/upgrade gradle

jvwilge:
Gradle is upgraded to a more recent version with the command `./gradlew wrapper --gradle-version 7.4.2`

The maven plugin is deprecated since gradle 7 and thus removed : https://docs.gradle.org/7.0/userguide/upgrading_version_6.html#removal_of_the_legacy_maven_plugin

Some `testCompile` steps were now giving errors, but they could be safely removed

Equalized the spring boot version because of an error with `2.1.1` : "property 'bootInf' is missing an input or output annotation"  https://github.com/spring-projects/spring-boot/issues/18326


stavshamir:
That's an unexpected and very nice contribution, thanks! I just have to check if the publishing plugins work correctly with the new version, I remember trying to upgrade to Gradle 6 and they didn't work properly.

jvwilge:
We were fiddling around in our team and I thought I might as wel create a PR with the improvements.

Local publishing gives the following error :
```
Unable to generate signature for '/Users/jeroen/Documents/springwolf-core/springwolf-core/build/generated-pom.xml' as no signatory is available to sign
```
But I think that is expected. 

Let me know whether I can also help out here. The project saved us so much time that I can spent some more on it 😄

stavshamir:
Everything works great 😃 And it's nice to hear it saved you time, thanks again for the PR

PR #75, opened 2022-06-19
Remove Project Title from Return Docs

Crain-32:
I have no idea if there needs to be more changed. 

Feel free to let me know if there is.


stavshamir:
There is no need to add the header. 
I added the title map thing to support multiple docs display in the ui, which no one uses at the moment anyway.
Before we release this change, the ui has to be changed to support single doc return body again.

timonback:
While looking into a way to lint the asyncApi docs, I came across the same issue.
I will address it in https://github.com/springwolf/springwolf-core/pull/96 including UI changes.

PR #77, opened 2022-06-21
Allow for manual consumer definition.

jrlambs:
Adds new ConsumerData classes for amqp and kafka.
allows for manual definition of the consumers using new ConsumerData classes.


stavshamir:
This looks quite good! 
I don't think it has to be any more complicated than this at this point.
Just make sure that you add unit tests and fix the existing integration tests and that all variable/function names has been changed from producer.
The documentation goes in the site repo.

PR #84, opened 2022-08-13
Add optional message description for manual consumer/producer definition

timonback:
Allow to set a optional description on the async api message object
spec: https://www.asyncapi.com/docs/reference/specification/v2.0.0#messageObject

Demo code for kafka and amqp examples was added as well

Related change in springwolf-ui will follow


timonback:
(FYI: I forked from HEAD-1, since the build on master is currently broken due to a change added in the last commit)

stavshamir:
> (FYI: I forked from HEAD-1, since the build on master is currently broken due to a change added in the last commit)

Yep, I will revert the last commit before merging this PR.

timonback:
I updated the PR so that also the springwolf-example projects tests pass - in my fork.

The core build is broken as the gh action builds the merged branch, not the changes only.

stavshamir:
Thanks for the nice addition @timonback :)

Issue #51, opened 2022-03-24
Support Kafka Header


sshemirani:
There should be a support to add Kafka headers. For example the spring __TypeId__ header should be provided in order for the spring to serialise and deserialize payloads. Without this header the publish button on the example tab cannot work as it need a __TypeId__ header to do its job. 

Thanks and keep up the good work.

stavshamir:
Thanks for the kind words!
I agree that this is a needed improvement. I'll try to add it in the next version.

sam0r040:
I decided to give it a try and started with the changes needed in the backend in PR https://github.com/springwolf/springwolf-core/pull/86

Let me know what you think... 

If we are happy with the backend implementation I would also try to provided the change for the frontend.

stavshamir:
Thanks for the great work of @sam0r040 and @timonback , this will be included in the next release 🥳

PR #86, opened 2022-08-26
Extend publishing api for kafka plugin to support publishing of kafka headers

sam0r040:
This PR should be a starting point and base for discussion to implement https://github.com/springwolf/springwolf-core/issues/51

The SpringwolfKafkaController now expects a structred request body, that contains a payload field and an optional header field to pass the kafka headers. The headers and the payload will be converted into a ProducerRecord and then passed on to the configured KafkaTemplate. I also considers the other two obvious options (http headers, query parameters) but somehow felt that this solution might be the most straight forward.

I also changed how SpringWolfKafkaProducer and its KafkaTemplate work together to improve testability and keep the KafkaTemplate out of an applications spring context.

### DIFF
--- a/springwolf-plugins/springwolf-kafka-plugin/src/main/java/io/github/stavshamir/springwolf/configuration/SpringwolfKafkaTemplateFactory.java
+++ b/springwolf-plugins/springwolf-kafka-plugin/src/main/java/io/github/stavshamir/springwolf/configuration/SpringwolfKafkaTemplateFactory.java
@@ -1,0 +1,45 @@
+package io.github.stavshamir.springwolf.configuration;
+
+import com.asyncapi.v2.model.server.Server;
+import com.google.common.collect.ImmutableMap;
+import lombok.RequiredArgsConstructor;
+import org.apache.kafka.clients.producer.ProducerConfig;
+import org.apache.kafka.common.serialization.StringSerializer;
+import org.springframework.kafka.core.DefaultKafkaProducerFactory;
+import org.springframework.kafka.core.KafkaTemplate;
+import org.springframework.kafka.support.serializer.JsonSerializer;
+import org.springframework.stereotype.Component;
+
+import java.util.Map;
+import java.util.Optional;
+
+@Component
+@RequiredArgsConstructor
+public class SpringwolfKafkaTemplateFactory {
+
+    private final AsyncApiDocket docket;
+
+    public Optional<KafkaTemplate<Object, Map<String, ?>>> buildKafkaTemplate() {
+        return getBootstrapServers(docket)
+                .map(this::buildProducerConfiguration)
+                .map(producerConfiguration -> new DefaultKafkaProducerFactory<Object, Map<String, ?>>(producerConfiguration))
+                .map(KafkaTemplate::new);
+    }
+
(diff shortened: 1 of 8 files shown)


stavshamir:
Thanks for the PR and sorry for the late response!

Unfortunately I did not have the time I would like to put into this project lately, but I will try to get to reviewing this in the coming weeks.

sam0r040:
No worries :). I would also try to add the headers to the client and generate header fields from Annotations, but I would definitely need some code pointers/ideas to start.

stavshamir:
Hi @sam0r040, I finally got to reviewing this (sorry it took so long)! This looks like a great addition. I don't have any comments code-wise, but I do have some questions:
1. The use case which raised this issue is producing a message for a topic with multiple possible payloads. Did you try to see if this PR solve this issue?
2. It can take place in another PR, but did you consider generating the type headers automatically (adding them to the async api doc)
3. We can also consider adding support for the header annotation for consumers (and manual consumer/producer) in a future PR.

Thanks for the nice work!

sam0r040:
We extended the ui to also send the headers in this PR https://github.com/springwolf/springwolf-ui/pull/26

springwolf-plugins/springwolf-kafka-plugin/src/main/java/io/github/stavshamir/springwolf/producer/SpringwolfKafkaProducer.java
```diff
@@ -1,72 +1,48 @@
...
+import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Collectors;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
 
 @Slf4j
-@Service
```
stavshamir, line 20: Is this on purpose? Without this it will not be injected to the controller, or am I missing something?

springwolf-plugins/springwolf-kafka-plugin/src/main/java/io/github/stavshamir/springwolf/configuration/SpringwolfKafkaTemplateFactory.java
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.kafka.core.KafkaTemplate;
+import org.springframework.kafka.support.serializer.JsonSerializer;
+import org.springframework.stereotype.Component;
+
+import java.util.Map;
+import java.util.Optional;
+
+@Component
+@RequiredArgsConstructor
+public class SpringwolfKafkaTemplateFactory {
```
stavshamir, line 18: I like this extraction from the producer class, makes much more sense, thanks!

springwolf-plugins/springwolf-kafka-plugin/src/main/java/io/github/stavshamir/springwolf/producer/SpringwolfKafkaProducer.java
sam0r040: The PoducerConfiguration acts as Beanfactory and creates the Producer (but I think you have seen it already).

Issue #83, opened 2022-07-11
Documentation is not correctly generated when having producer and consumer for the same topic. 


danielatchobanikova:
**Describe the bug**
1. We have a Spring Boot microservice, let's say it is called "test-service"
2. "test-service" has the following consumers:
     2.1 ConsumerA -> consumers "topic-a"
     2.2 ConsumerB -> consumer "topic-b"
3. Now let's say we configure an asyncApiDocket Bean. Everything works fine and I can see both consumers in springwolf ui.
4. Now we add ProducerA -> produces events to "topic-a'.
5. Now springwolf ui display only consumer for topic-b and producer for topic-a. Seems the consumer for topic-a documentation was overwritted.

**Dependencies and versions used**
`springwolf-kafka` version `0.6.1`. I tried with `0.7.0` as well.

**Code example**
```
@SpringBootApplication(scanBasePackages = {})

public class ExampleApplication {

    public static void main(String[] args) {

        SpringApplication.run(ExampleApplication.class, args);
    }
}

```

```
@Component
@RequiredArgsConstructor
@Slf4j
@KafkaListener(
    batch = "true",
    topics = "topic-a"
)
public class ConsumerA {
    @KafkaHandler
    public void consume(Object event) {
    }
}
```

```
@Component
@RequiredArgsConstructor
@Slf4j
@KafkaListener(
    batch = "true",
    topics = "topic-b"
)
public class ConsumerB {
    @KafkaHandler
    public void consume(Object event) {
    }
}
```

```
@EnableAsyncApi
@Configuration
public class AsyncApiConfigurationTest {

    private static final String LISTENERS_SCAN_PACKAGE =
        "example.consumer";

    @Bean
    public AsyncApiDocket asyncApiDocket(
        @Value("${service.name}") String appName,
        @Value("${service.version}") String appVersion,
        @Value("${spring.kafka.bootstrap-servers}") String bootstrapServers) {

        Info info = Info.builder()
            .version(appVersion)
            .title(appName)
            .description("description")
            .build();

        String producerTopicName = "topic-a";
        ProducerData preparedDataProducer = ProducerData.builder()
            .channelName(producerTopicName)
            .operationBinding(ImmutableMap.of("kafka", new KafkaOperationBinding()))
            .payloadType(FetchUserBalances.class)
            .build();

        return AsyncApiDocket.builder()
            .basePackage(LISTENERS_SCAN_PACKAGE)
[...]

PR #88, opened 2022-09-07
GH-83: fix: Merge operations when repeated ChannelItem found

timonback:
As mentioned in GH-83, only one channelItem is part of the final map - although both channelItems are found.

This improves the situation by merging the different operations for the same topic name into the same ChannelItem - while still the last one wins. This might require further improvement in the future.


stavshamir:
That's a great solution for that bug, thanks again!

Issue #87, opened 2022-09-01
Payload schema is not documented in a batch kafka listener 


bdridi:
**Describe the bug**

For a Kafka batch listener where a the type  of the the parameter of the listener method is `List<ExamplePayload>`, the schema of `ExamplePayload` is not documented. The consumer is displayed with a schema : `java.util.List` 
 
**Dependencies and versions used**

For example, `springwolf-kafka` version `0.7.0

**Code example**

```kotlin
 @KafkaListener(
        topics = ["topic-example"],
        containerFactory = "ExamplePayloadKafkaListenerContainerFactory"
    )
    fun examplePayloadBatchListener(examplePayloadList: List<ExamplePayload?>) {
        // ...
    }
```

stavshamir:
Since type erasure it is challenging to get the type parameters at run time. I guess there are ways to overcome this, but the solution is usually on the client side - to have a class wrapping the generic type. 

If you can come up with a better solution please let me know.

stavshamir:
Done thanks to @timonback . Will be available in the next release.

PR #89, opened 2022-09-08
feat: Extract generic type for KafkaListeners in batch mode

timonback:
The KafkaPayloadTypeExtractor contains the logic to extract the payload type, which previously was part of ClassLevelKafkaListenerScanner and MethodLevelKafkaListenerScanner.

If the parameter is of type List<?>, reflection is used to extract the underlying type.
This solution assumes by using the List type, Kafka batch mode is used.

A future improvement can be to only extract the generic when the KafkaListener annotation is used together with the batch property. A library update for spring-kafka is possible: https://docs.spring.io/spring-kafka/api/org/springframework/kafka/annotation/KafkaListener.html#batch()

To demo it, one listener in ExampleConsumer is updated to use batch mode.

Relates to https://github.com/springwolf/springwolf-core/issues/87


springwolf-plugins/springwolf-kafka-plugin/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/KafkaPayloadTypeExtractor.java
```diff
@@ -0,0 +1,78 @@
...
+    static Class<?> getPayloadParameterClass(Method method, Class<?>[] parameterTypes, int parameterPayloadIndex) {
+        Class<?> parameterClass = parameterTypes[parameterPayloadIndex];
+
+        try {
+            // Resolve generic type for kafka batch listeners
+            if (parameterClass == List.class) {
+                Type type = ((ParameterizedTypeImpl) method.getGenericParameterTypes()[parameterPayloadIndex]).getActualTypeArguments()[0];
+                parameterClass = Class.forName(type.getTypeName());
+            }
+        } catch (ClassNotFoundException ex) {
```
stavshamir, line 37: I think the catch should be for `Exception`, it looks like `getActualTypeArguments` can also throw exceptions and may return an empty array. See: https://docs.oracle.com/javase/7/docs/api/java/lang/reflect/ParameterizedType.html

springwolf-plugins/springwolf-kafka-plugin/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/KafkaPayloadTypeExtractor.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    static Class<?> getPayloadParameterClass(Method method, Class<?>[] parameterTypes, int parameterPayloadIndex) {
+        Class<?> parameterClass = parameterTypes[parameterPayloadIndex];
+
+        try {
+            // Resolve generic type for kafka batch listeners
+            if (parameterClass == List.class) {
+                Type type = ((ParameterizedTypeImpl) method.getGenericParameterTypes()[parameterPayloadIndex]).getActualTypeArguments()[0];
+                parameterClass = Class.forName(type.getTypeName());
```
stavshamir, line 35: It makes more sense to me to simply return here instead of assigning and returning later. What do you think?

springwolf-plugins/springwolf-kafka-plugin/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/MethodLevelKafkaListenerScanner.java
```diff
@@ -72,46 +70,6 @@ protected String getChannelName(KafkaListener annotation) {
 
     @Override
     protected Class<?> getPayloadType(Method method) {
-        String methodName = String.format("%s::%s", method.getDeclaringClass().getSimpleName(), method.getName());
```
stavshamir, line 75: It's always nice to see so much code getting deleted :)

springwolf-plugins/springwolf-kafka-plugin/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/KafkaPayloadTypeExtractor.java
```diff
@@ -0,0 +1,78 @@
...
+import sun.reflect.generics.reflectiveObjects.ParameterizedTypeImpl;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
+import java.lang.reflect.Type;
+import java.util.Arrays;
+import java.util.List;
+
+@Slf4j
+public class KafkaPayloadTypeExtractor {
```
stavshamir, line 14: The ampq plugin also has this exact code. Maybe we should catch this opportunity to use this new class for the amqp scanner as well? The payload type can be a list there as well (and also for non-batch kafka,=)

stavshamir (commented):
Great work, it's good to see so much code re-used. Please see my comments, I really think we can use this extractor for the amqp plugin as well

springwolf-plugins/springwolf-kafka-plugin/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/KafkaPayloadTypeExtractor.java
timonback: Done

springwolf-plugins/springwolf-kafka-plugin/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/KafkaPayloadTypeExtractor.java
timonback: Sure, I updated it. It's a matter of which style you prefer (return early vs return only once).

springwolf-plugins/springwolf-kafka-plugin/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/KafkaPayloadTypeExtractor.java
timonback: The payload extractor was moved to core now (requiring the spring-messaging dependency in core as well)

timonback:
I updated the PR with a larger refactoring and moving duplicated code to Abstract classes in core. The amqp and kafka listeners are slimmer now, although there is still some complexity through the abstract methods.
For reviewing, I suggest to check out the code locally.

Since the release is around the corner, I think this is a feature for the following release.

Regarding the batch mode, yes, this by default assumes that all List<?> Payloads are batch methods. In the future this could be made configurable (batch-mode: yes/no), but at this point I would assume that people define their own payload types and not use List<?> in non-batch mode (defining `<MyType extends List<?>` should work around the list detection anyway).

stavshamir:
I really like these refactoring, it will make it much much easier to implement more plugins based on listener annotations. Great job!

PR #92, opened 2022-09-16
test: Run ApiIntegrationTests with embeddedKafka

timonback:
For some reason, I am not able to get the integration test running locally with DockerComposeContainer. Running the docker-compose file manually poses no issue.

This runs the same test, however with an embeddedkafka.
As this does not use the docker container of the application anymore, this test functions differently and to release the docker image, the image must be build explicitly.

If this change is fine, I will add the same setup for amqp/rabbitmq. I can also keep this in my fork, if this breaks your workflow.


stavshamir:
I guess it will help contributors who are less familiar with docker to contribute to the project. 
It does not break the flow, I just wonder if we should leave the docker compose tests as well. I also wonder if anybody uses it at all - I wanted to provide a quick peek for potential users, but I don't really know if people use it at all. I can see the download statistics on the docker image but it may well be the CI downloading them.

timonback:
I noticed, that the docker image is build locally and tests are run against it.
Therefore, I decided to add keep the ApiIntegrationTests as ApiIntegrationWithDockerTests. For people who are unable to run them, I added the note regarding disabeling them locally.

I also just added the same for amqp.

Issue #94, opened 2022-09-16
Add support for kafka headers


timonback:
Springwolf already provides documentation for the data payload. Transport metadata (aka headers) cannot be documented with springwolf at this point.
Why the separation? The data/domain payload should be separate from the transport metadata. In REST APIs, the HTTP Content-Type, Content-Length is also a header and (usually) not part of the response object.

AsyncApi allows to define headers as part of the message, similar as the payload is defined: https://www.asyncapi.com/docs/reference/specification/v2.4.0#messageObject

Personal use-case:
https://cloudevents.io is an upcoming standardization, which relies on putting metadata into headers. I would like to document those headers. I do not intend to create any kind of dependency on cloudevents.

stavshamir:
See #86

sam0r040:
Just to keep everything together, the issue that triggered me for PR https://github.com/springwolf/springwolf-core/pull/86 : Issue https://github.com/springwolf/springwolf-core/issues/51

timonback:
By reading #51, I was not sure if documenting the headers is part of it, or only making the springwolf-ui produce to topic functionality work.
If documenting the headers as part of springwolf-ui is part of it, this is a duplicate and can be closed.

stavshamir:
That's a good point, I was just commenting about it in #51. @sam0r040 in his PR only provides a fix for producing to a topic which has multiple payload types, so let's leave this one (maybe describe it with more detail).

PR #99, opened 2022-09-30
feat: Add possibility to manually document Kafka header for consumers and producers

timonback:
Co-authored-by: sam0r040 <93372330+sam0r040@users.noreply.github.com>

Closes: https://github.com/springwolf/springwolf-core/issues/94


springwolf-examples/springwolf-kafka-example/src/main/java/io/github/stavshamir/springwolf/example/configuration/AsyncApiConfiguration.java
```diff
@@ -35,12 +39,14 @@ public AsyncApiDocket asyncApiDocket() {
         KafkaProducerData exampleProducerData = KafkaProducerData.kafkaProducerDataBuilder()
                 .topicName(PRODUCER_TOPIC)
                 .payloadType(ExamplePayloadDto.class)
+                .headers(createSpringDefaultHeaders())
```
timonback, line 42: Usage is outlined here

springwolf-examples/springwolf-kafka-example/src/main/java/io/github/stavshamir/springwolf/example/configuration/AsyncApiConfiguration.java
```diff
@@ -59,4 +65,59 @@ public AsyncApiDocket asyncApiDocket() {
                 .build();
     }
 
+    private static AsyncHeaders createSpringDefaultHeaders() {
```
stavshamir, line 68: We should consider providing factory methods for common uses like spring type id header and cloud events in `AsyncHeaders`

stavshamir (commented):
Amazing job, this would be really useful! 
I am trying to understand the defaults here - when is `UKNOWN` being used and when `NONE`?

springwolf-examples/springwolf-kafka-example/src/main/java/io/github/stavshamir/springwolf/example/configuration/AsyncApiConfiguration.java
sam0r040: We also thought that this might be a good idea and updated the PR accordingly. 
The builder for Spring headers are in the spring kafka extension.  
The builder for Cloud Events is located in core, because Cloud Event headers are more general and not limited to Kafka (or another messaging system) as long as headers are supported.
The Kafka example has been updated accordingly.

stavshamir:
Now we just need to make the spring type headers generate automatically :)

PR #104, opened 2022-10-21
Auto detect spring kafka headers

timonback:


springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/types/channel/operation/message/PayloadReference.java
```diff
@@ -3,12 +3,16 @@
...
 
 @NoArgsConstructor
 @EqualsAndHashCode
+@ToString
 public class PayloadReference {
 
     @Getter
+    @EqualsAndHashCode.Include
+    @ToString.Include
     private String $ref;
```
timonback, line 16: FYI: lombok EqualsAndHashCode and ToString ignore fields with a `$` prefix.
Following, also some test a fixed.

stavshamir: Good to know, thanks

stavshamir:
Great, can't wait to see it working :)

PR #113, opened 2022-10-25
chore/update gradle

stavshamir:


timonback (approved):
LGTM
Great simplification of the signing configuration.

Also, building locally will become so much easier as signing isnt done on the build task target anymore, but only publication

PR #118, opened 2022-11-25
Annotation for publisher and subscriber

timonback:
In order to simplify documenting producers, we added annotations to document producers where the actual send methods are being called.
This is an additional way to define `ProducerData` and `ConsumerData`.

The following example is provided in the code as well (kafka + amqp):
```
@AsyncPublisher(operation = @AsyncOperation(
            channelName = PRODUCER_TOPIC,
            description = "Custom, optional description defined in the AsyncPublisher annotation",
            headers = @AsyncOperation.Headers(
                    schemaName = "SpringKafkaDefaultHeaders",
                    values = {
                            @AsyncOperation.Headers.Header(
                                    name = DEFAULT_CLASSID_FIELD_NAME,
                                    description = "Spring Type Id Header",
                                    value = "io.github.stavshamir.springwolf.example.dtos.ExamplePayloadDto"
                            ),
                            @AsyncOperation.Headers.Header(
                                    name = DEFAULT_CLASSID_FIELD_NAME,
                                    description = "Spring Type Id Header",
                                    value = "io.github.stavshamir.springwolf.example.dtos.AnotherPayloadDto"
                            ),
                    }
            )
    ))
    @KafkaAsyncOperationBinding(
            bindingVersion = "1",
            clientId = "foo-clientId",
            groupId = "foo-groupId"
    )
```

The `KafkaAsyncOperationBinding` annotation is optional, as are most fields that are not required (like the description and headers).

Review the changes in the example project for more details.

In terms of implementation, additional scanners have been defined, as well as the annotations themselves. As multiple ChannelItem might be found for a channel now (spring listener annotation + AsyncConsumer + ConsumerData in the Docket), the Merger builds the final channel item. By specifying priorities, custom annotations beat auto discovered definitions.

### DIFF
--- a/springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/ChannelMerger.java
+++ b/springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/ChannelMerger.java
@@ -6,7 +6,6 @@ import io.github.stavshamir.springwolf.asyncapi.MessageHelper;
 import io.github.stavshamir.springwolf.asyncapi.types.channel.operation.message.Message;
 
 import java.util.*;
-import java.util.function.Function;
 
 import static io.github.stavshamir.springwolf.asyncapi.MessageHelper.toMessageObjectOrComposition;
 
@@ -34,29 +33,29 @@ public class ChannelMerger {
                 mergedChannels.put(entry.getKey(), entry.getValue());
             } else {
                 ChannelItem channelItem = mergedChannels.get(entry.getKey());
-                channelItem.setPublish(mergeOperation(channelItem, entry.getValue(), ChannelItem::getPublish));
-                channelItem.setSubscribe(mergeOperation(channelItem, entry.getValue(), ChannelItem::getSubscribe));
+                channelItem.setPublish(mergeOperation(channelItem.getPublish(), entry.getValue().getPublish()));
+                channelItem.setSubscribe(mergeOperation(channelItem.getSubscribe(), entry.getValue().getSubscribe()));
             }
         }
 
         return mergedChannels;
     }
 
-    private static Operation mergeOperation(ChannelItem channelItem, ChannelItem otherChannelItem, Function<ChannelItem, Operation> getOperation) {
-        Set<Message> mergedMessages = getChannelMessages(channelItem, getOperation);
-        Set<Message> currentEntryMessages = getChannelMessages(otherChannelItem, getOperation);
+    private static Operation mergeOperation(Operation operation, Operation otherOperation) {
+        Set<Message> mergedMessages = getMessages(operation);
+        Set<Message> currentEntryMessages = getMessages(otherOperation);
         mergedMessages.addAll(currentEntryMessages);
 
-        Operation operation = getOperation.apply(channelItem) != null ? getOperation.apply(channelItem) : getOperation.apply(otherChannelItem);
+        Operation mergedOperation = operation != null ? operation : otherOperation;
         if (!mergedMessages.isEmpty()) {
-            operation.setMessage(toMessageObjectOrComposition(mergedMessages));
+            mergedOperation.setMessage(toMessageObjectOrComposition(mergedMessages));
         }
-        return operation;
+        return mergedOperation;
     }
 
-    private static Set<Message> getChannelMessages(ChannelItem channelItem, Function<ChannelItem, Operation> getOperation) {
+    private static Set<Message> getMessages(Operation operation) {
         return Optional
-                .ofNullable(getOperation.apply(channelItem))
+                .ofNullable(operation)
                 .map(Operation::getMessage)
                 .map(MessageHelper::messageObjectToSet)
                 .orElseGet(TreeSet::new);
--- a/springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/ChannelMerger.java
+++ b/springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/ChannelMerger.java
@@ -1,0 +1,50 @@
+package io.github.stavshamir.springwolf.asyncapi.scanners.channels;
+
+import com.asyncapi.v2.model.channel.ChannelItem;
+import com.asyncapi.v2.model.channel.operation.Operation;
+import io.github.stavshamir.springwolf.asyncapi.MessageHelper;
+import io.github.stavshamir.springwolf.asyncapi.types.channel.operation.message.Message;
+
+import java.util.*;
+import java.util.function.Function;
+
+import static io.github.stavshamir.springwolf.asyncapi.MessageHelper.toMessageObjectOrComposition;
+
+public class ChannelMerger {
+
+    public static Map<String, ChannelItem> merge(List<Map.Entry<String, ChannelItem>> channelEntries) {
+        Map<String, ChannelItem> mergedChannels = new TreeMap<>();
+
+        for (Map.Entry<String, ChannelItem> entry : channelEntries) {
+            if (!mergedChannels.containsKey(entry.getKey())) {
+                mergedChannels.put(entry.getKey(), entry.getValue());
+            } else {
+                ChannelItem channelItem = mergedChannels.get(entry.getKey());
+                channelItem.setPublish(mergeOperation(channelItem, entry.getValue(), ChannelItem::getPublish));
+                channelItem.setSubscribe(mergeOperation(channelItem, entry.getValue(), ChannelItem::getSubscribe));
+            }
+        }
+
+        return mergedChannels;
+    }
+
+    private static Operation mergeOperation(ChannelItem channelItem, ChannelItem otherChannelItem, Function<ChannelItem, Operation> getOperation) {
+        Set<Message> mergedMessages = getChannelMessages(channelItem, getOperation);
+        Set<Message> currentEntryMessages = getChannelMessages(otherChannelItem, getOperation);
+        mergedMessages.addAll(currentEntryMessages);
+
+        Operation operation = getOperation.apply(channelItem) != null ? getOperation.apply(channelItem) : getOperation.apply(otherChannelItem);
+        if (!mergedMessages.isEmpty()) {
+            operation.setMessage(toMessageObjectOrComposition(mergedMessages));
+        }
+        return operation;
+    }
+
+    private static Set<Message> getChannelMessages(ChannelItem channelItem, Function<ChannelItem, Operation> getOperation) {
+        return Optional
+                .ofNullable(getOperation.apply(channelItem))
+                .map(Operation::getMessage)
+                .map(MessageHelper::messageObjectToSet)
+                .orElseGet(TreeSet::new);
+    }
+}
--- a/springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/ChannelMerger.java
+++ b/springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/ChannelMerger.java
@@ -10,8 +10,22 @@ import java.util.function.Function;
 
 import static io.github.stavshamir.springwolf.asyncapi.MessageHelper.toMessageObjectOrComposition;
 
+
+/**
+ * Util to merge multiple {@link ChannelItem}s
+ */
 public class ChannelMerger {
 
+    /**
+     * Merges multiple channelItems by channel name
+     * <p>
+     * Given two channelItems for the same channel name, the first seen ChannelItem is used
+     * If an operation is null, the next non-null operation is used
+     * Messages within operations are merged
+     *
+     * @param channelEntries Ordered pairs of channel name to ChannelItem
+     * @return A map of channelName to a single ChannelItem
+     */
     public static Map<String, ChannelItem> merge(List<Map.Entry<String, ChannelItem>> channelEntries) {
         Map<String, ChannelItem> mergedChannels = new TreeMap<>();
 
--- a/springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/ChannelPriority.java
+++ b/springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/ChannelPriority.java
@@ -1,0 +1,28 @@
+package io.github.stavshamir.springwolf.asyncapi.scanners.channels;
+
+import io.github.stavshamir.springwolf.asyncapi.scanners.channels.operationdata.annotation.AsyncPublisher;
+import io.github.stavshamir.springwolf.asyncapi.scanners.channels.operationdata.annotation.AsyncSubscriber;
+import io.github.stavshamir.springwolf.configuration.AsyncApiDocket;
+
+public class ChannelPriority {
+    /**
+     * Manual defined channels have the highest priority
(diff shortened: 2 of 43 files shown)


stavshamir:
Thanks @timonback and @sam0r040 , this will be a big improvement. A lot of content so may take me a while to review, but I'm on it

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/serializers/KafkaOperationBindingSerializer.java
```diff
@@ -33,8 +33,7 @@ private void writeGroupId(KafkaOperationBinding value, JsonGenerator gen) throws
         gen.writeFieldName("groupId");
         gen.writeStartObject();
         gen.writeStringField("type", "string");
-        gen.writeFieldName("enum");
-        gen.writeStartArray();
+        gen.writeArrayFieldStart("enum");
```
stavshamir, line 36: Nice :)

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/types/channel/operation/message/Message.java
```diff
@@ -35,4 +34,13 @@ public class Message {
     private PayloadReference payload;
 
     private HeaderReference headers;
+
+
+    @Override
+    public int compareTo(Message o) {
```
stavshamir, line 40: Why? For sorting?

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/annotation/AsyncPublisher.java
```diff
@@ -0,0 +1,21 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation is mapped to {@link ProducerData}
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(value = {ElementType.METHOD})
+public @interface AsyncPublisher {
```
stavshamir, line 16: What I'm missing here, when I'm taking the user POV, is where can I use this? Can every method be annotated by this? If not, what is the signature structure/pattern that is allowed? 
It looks like the payload type can be extracted if the method has a certain signature. If it doesn't, the javadoc should explain how to specify the payload type explicitly, etc.

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/annotation/AsyncSubscriber.java
```diff
@@ -0,0 +1,21 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation is mapped to {@link ConsumerData}
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(value = {ElementType.METHOD})
+public @interface AsyncSubscriber {
```
stavshamir, line 16: What I'm missing here, when I'm taking the user POV, is where can I use this? Can every method be annotated by this? If not, what is the signature structure/pattern that is allowed? 
It looks like the payload type can be extracted if the method has a certain signature. If it doesn't, the javadoc should explain how to specify the payload type explicitly, etc.

springwolf-examples/springwolf-kafka-example/src/main/java/io/github/stavshamir/springwolf/example/producers/ExampleProducer.java
```diff
@@ -1,18 +1,46 @@
...
+                                    description = "Spring Type Id Header",
+                                    value = "io.github.stavshamir.springwolf.example.dtos.AnotherPayloadDto"
+                            ),
+                    }
+            )
+    ))
+    @KafkaAsyncOperationBinding(
+            bindingVersion = "1",
+            clientId = "foo-clientId",
+            groupId = "foo-groupId"
```
stavshamir, line 42: I'm pretty sure group Id is only relevant for consumers. If so let's remove this attribute from the producers' examples.

springwolf-plugins/springwolf-kafka-plugin/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/annotation/KafkaAsyncOperationBinding.java
```diff
@@ -0,0 +1,19 @@
+package io.github.stavshamir.springwolf.asyncapi.scanners.channels.operationdata.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(value={ElementType.METHOD})
+public @interface KafkaAsyncOperationBinding {
```
stavshamir, line 10: As with the other annotations, we should include some info about how and when to use this

springwolf-plugins/springwolf-amqp-plugin/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/annotation/AmqpAsyncOperationBinding.java
```diff
@@ -0,0 +1,27 @@
+package io.github.stavshamir.springwolf.asyncapi.scanners.channels.operationdata.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(value = {ElementType.METHOD})
+public @interface AmqpAsyncOperationBinding {
```
stavshamir, line 10: As with the other annotations, we should include some info about how and when to use this

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/ChannelPriority.java
```diff
@@ -0,0 +1,28 @@
+package io.github.stavshamir.springwolf.asyncapi.scanners.channels;
+
+import io.github.stavshamir.springwolf.asyncapi.scanners.channels.operationdata.annotation.AsyncPublisher;
+import io.github.stavshamir.springwolf.asyncapi.scanners.channels.operationdata.annotation.AsyncSubscriber;
+import io.github.stavshamir.springwolf.configuration.AsyncApiDocket;
+
+public class ChannelPriority {
```
stavshamir, line 7: Very nice idea!

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/annotation/AsyncPublisherAnnotationScanner.java
```diff
@@ -0,0 +1,88 @@
...
+
+        return Arrays.stream(type.getDeclaredMethods())
+                .filter(method -> method.isAnnotationPresent(annotationClass))
+                .collect(toSet());
+    }
+
+    private OperationData mapMethodToOperationData(Method method) {
+        log.debug("Mapping method \"{}\" to channels", method.getName());
+
+        Map<String, OperationBinding> operationBindings = AsyncAnnotationScannerUtil.processBindingFromAnnotation(method, operationBindingProcessors);
```
stavshamir, line 66: I might be missing something, but what happens if not operation binding annotations has been used?

stavshamir (commented):
Great PR again guys!
I like the ideas you present here, I think we should improve javadocs of the annotations to help the users understand how to use them.

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/ChannelMerger.java
```diff
@@ -0,0 +1,64 @@
...
+                ChannelItem channelItem = mergedChannels.get(entry.getKey());
+                channelItem.setPublish(mergeOperation(channelItem, entry.getValue(), ChannelItem::getPublish));
+                channelItem.setSubscribe(mergeOperation(channelItem, entry.getValue(), ChannelItem::getSubscribe));
+            }
+        }
+
+        return mergedChannels;
+    }
+
+    private static Operation mergeOperation(ChannelItem channelItem, ChannelItem otherChannelItem, Function<ChannelItem, Operation> getOperation) {
```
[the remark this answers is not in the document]
timonback: True, that is indeed easier.
Good catch

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/annotation/AsyncAnnotationScannerUtil.java
```diff
@@ -0,0 +1,54 @@
...
+                                    .enumValue(values)
+                                    .example(values.stream().findFirst().orElse(null))
+                                    .build()
+                    );
+                });
+
+        return headers;
+    }
+
+    public static String nullIfEmpty(String stringValue) {
```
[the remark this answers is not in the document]
timonback: Good catch

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/annotation/AsyncPublisherAnnotationScanner.java
timonback: If no operation binding is used, also none is added.
No exception is thrown.

It is also possible to add multiple bindings to the same method (amqp + kafka) - if that makes sense in the application

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/types/channel/operation/message/Message.java
timonback: The ChannelMerger#mergeOperation uses a Set to remove duplicate messages for the same channel.
The "same" message can be provided by multiple scanners, but with slightly different title, description, header information.

If you find it unsuitable here, I can also look into alternatives to detect "duplicates"

stavshamir: In that case:
1. Wouldn't the equals and hashCode be enough? Why is Comparable required?
2. I think that in case of multiple messages in an operation, the differentiator should be the payload type, no?

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/annotation/AsyncPublisherAnnotationScanner.java
stavshamir: Makes sense, good design

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/types/channel/operation/message/Message.java
timonback: I had a closer look. What I described above is already achieved with the added `@EqualsAndHashCode(of = {"name"})`.

The field name is the canonical name of the class. With multiple messages, all messages with the same canonical name (=payload type) are merged together.

I added one more test to the ChannelMergerTest to confirm this. This also confirms, that the `compareTo` is not needed and therefore removed in the latest commit.

timonback:
Thanks for all the feedback. I have updated the PR accordingly.

pdalfarr:
Hi @timonback ,

I have one question about `@AsyncPublisher` usage. 

As far as I understand, the springwolf UI allows to send messages to QUEUES (when `plugin.amqp.publishing.enabled: true`  )
which is OK when specifying queue name in AsyncOperation.channelName

But I do not understand how to send a message to a TOPIC.

Let's take an example:

In one of your RabbitMQ your example, here

https://github.com/springwolf/springwolf-core/blob/master/springwolf-examples/springwolf-amqp-example/src/main/java/io/github/stavshamir/springwolf/example/amqp/producers/ExampleProducer.java#L17-L28

we have:

```
   @AsyncPublisher(
            operation =
                    @AsyncOperation(
                            channelName = "example-producer-channel-publisher",
                            description = "Custom, optional description defined in the AsyncPublisher annotation"))
    @AmqpAsyncOperationBinding()
    public void sendMessage(ExamplePayloadDto msg) {
        // send
        AnotherPayloadDto dto = new AnotherPayloadDto("fooValue", msg);
        rabbitTemplate.convertAndSend("example-topic-exchange", "example-topic-routing-key", dto);
    }
```

Now, let's suppose we have, in another micro-service, a method with the followinf annotation:

```
    @Override
    @RabbitListener(
            bindings = @QueueBinding(
                    value = @Queue(
                            name = "microservice-2-private-queue-name" // this is a private queu of this microservice: I do not want this string value to be known by the first microservice , i.e., your sample code
                    ),
                    exchange = @Exchange(
                            name = "example-producer-channel-publisher", // the exchange from in your sample code
                            type = ExchangeTypes.TOPIC
                    ),
                    key = "example-topic-routing-key" // the key from your sample code
            )
    )
```

So, this does declare a topic which receives messages send by 'your sample' code :

`   rabbitTemplate.convertAndSend("example-topic-exchange", "example-topic-routing-key", dto);
`

This is (hopefully) working fine, BUT, when using the Springwolf UI, it does not work.

It seems that messages sent from the UI goes to a QUEUE named "example-producer-channel-publisher" along with a routing-key with the same value, i.e.  "example-producer-channel-publisher"

**Question:**
Is it a way to use AsyncOperation to express the fact that we want 'the UI' to send message towards a given exchange (a topic in my example), along with a routing key ? 

In the example I gave here, your code (first microservice) is sending a message to a topic (rabbitTemplate.convertAndSend("example-topic-exchange", "example-topic-routing-key", dto);
).
Of course, I do NOT the "microservice-2-private-queue-name" queue name (from micro-service 2) to be known by the AsyncPublisher annotation of micro-service 1.

And I want my second micro service being able to change 'at will' the name of it's own internal/private queue.
microservice one should only know about the exchange+routing-key of microservice 2. In other words, microservice one, should send message to this exchange (and this is what is actually done in the code)
So I guess this should be reflected in the annotation.

I tested this on my side in my project, and the messages send from the UI get delivered to microservice 2 if I set, in microservice 1, the channelName to "microservice-2-private-queue-name"

I hope my explanation is clear.

Thans for your feedback on that.

Pascal

timonback:
Hi @pdalfarr ,

I understand your feature request and agree that it is currently not supported and might require a specific amqpchannelbinding annotation or therelike.

I am not very familiar with rabbitmq and we see here that not all protocol details can be abstracted into one generic asynclistener annotation.

Feel free to open an enhancement issue and contribute it to the project.

PS: Commenting on closed issues is difficult to see for us.

PR #124, opened 2022-12-17
chore: Run spectral from docker container directly

timonback:
Avoid gh action that builds the spectral docker container first

Relates to https://github.com/stoplightio/spectral-action/issues/452#issuecomment-1285886971


stavshamir:
Good catch, let's be on the lookout for when they fix it

PR #125, opened 2022-12-17
Upgrade to spring boot 2.7.6, junit 5, gradle 7.5.1

timonback:
Also upgrade other dependencies
- lombok
- assertj-core
- swagger-core

Remove kotlin (converted test to java)

Relates to: https://github.com/springwolf/springwolf-core/issues/112
Relates to: https://github.com/springwolf/springwolf-core/issues/110


springwolf-plugins/springwolf-kafka-plugin/src/main/java/io/github/stavshamir/springwolf/asyncapi/types/channel/operation/message/header/AsyncHeadersForSpringKafkaBuilder.java
```diff
@@ -1,6 +1,6 @@
 package io.github.stavshamir.springwolf.asyncapi.types.channel.operation.message.header;
 
-import org.springframework.kafka.support.converter.AbstractJavaTypeMapper;
+import org.springframework.kafka.support.mapping.AbstractJavaTypeMapper;
```
timonback, line 3: Due to the spring update, also spring-kafka is updated, which includes this change.

If we are assuming that springwolf users are up-to-date, then this change is good. Otherwise, this might force users to upgrade spring-kafka.

I am not sure if we want to include in the next release as it contains already so many changes: async annotations + spring properties + spring-cloud + fixes.

Or, this could be a minor update right after the upcoming release (1.1.0)

springwolf-core/src/test/java/io/github/stavshamir/springwolf/asyncapi/DefaultAsyncApiSerializerServiceTest.java
```diff
@@ -18,20 +18,20 @@
...
+import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.test.context.ContextConfiguration;
-import org.springframework.test.context.junit4.SpringRunner;
 
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.util.Map;
 
-@RunWith(SpringRunner.class)
```
stavshamir, line 34: The JUnit5 equivalent would be `@ExtendWith(SpringExtension.class)`. `@SpringBootTest` would work as well, but will make the test much slower

springwolf-examples/springwolf-amqp-example/src/test/java/io/github/stavshamir/springwolf/example/SpringContextTest.java
```diff
@@ -1,23 +1,20 @@
...
 import org.springframework.test.context.TestPropertySource;
-import org.springframework.test.context.junit4.SpringRunner;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.Assert.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 
-@RunWith(Enclosed.class)
+@Nested
```
stavshamir, line 15: I thought `@Nested` only has meaning in inner classes. What does it do when it's top level?

stavshamir (commented):
Great work on this, feels good to get rid of JUnit4 and the kotlin tests!
Only one important comment - replacing JUnit4 `@RunWith(SpringRunner.class)` with `@SpringBootTest` will work but make tests much slower since Spring will load the entire context. Except for context and integration test, we should use `@ExtendWith(SpringExtension.class)` instead.

PR #129, opened 2023-01-10
Fix: Remove that EqualsAndHashCode of Message only uses the name, Cha…

sam0r040:
…nge MessageHelper to not expose the TreeSet that is used for deduplication because a TreeSet breaks equals

### DIFF
--- a/springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/MessageHelper.java
+++ b/springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/MessageHelper.java
@@ -4,7 +4,13 @@ import com.google.common.collect.ImmutableMap;
 import io.github.stavshamir.springwolf.asyncapi.types.channel.operation.message.Message;
 import lombok.extern.slf4j.Slf4j;
 
-import java.util.*;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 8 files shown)


springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/MessageHelper.java
```diff
@@ -13,6 +19,10 @@ public class MessageHelper {
     private static final String ONE_OF = "oneOf";
 
     private static final Comparator<Message> byMessageName = Comparator.comparing(Message::getName);
+
+    // TODO Why do we need a SortedSet here? Using a comparator with only the message name will break deep equals on the Set
```
sam0r040, line 23: I'm not sure of having a sorted set here is a requirement, but using the TreeSet for Sorting and deduplication causes issues when comparing (ie test do not break). Maybe there is a better solution than just using a hashset (looses ordering), but for the tests it seems to be ok ...

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/types/channel/operation/message/Message.java
```diff
@@ -10,7 +10,8 @@
  */
 @Data
 @Builder
-@EqualsAndHashCode(of = {"name"})
+// TODO Why ignore other fields?
+//@EqualsAndHashCode(of = {"name"})
```
sam0r040, line 14: I'm not sure what the motivation of changing equals and hashcode was, but I'd rather do a custom equals (based on name) where it is needed instead of changing it for all other code because this causes other issues (ie test don't break)

springwolf-core/src/test/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/annotation/AsyncPublisherAnnotationScannerTest.java
```diff
@@ -66,7 +66,7 @@ public void scan_componentHasPublisherMethod() {
         Message message = Message.builder()
                 .name(SimpleFoo.class.getName())
                 .title(SimpleFoo.class.getSimpleName())
-                .description(null)
+                .description("")
```
sam0r040, line 69: One of the issues that didn't come up because of the equals, the default value for the description is empty string

sam0r040:
Sorry, I accidentally created the PR here although I wanted to create it on my fork while im still developing...

stavshamir:
No problem, nice to see more stuff incoming :)

Issue #132, opened 2023-01-19
IllegalAccessError with springwolf-core 0.10.0 and kafka batch listener


timonback:
**Describe the bug**
Unable to start springwolf-core 0.10.0 with Kafka BatchListener

**Dependencies and versions used**
springwolf-core 0.10.0

**Code example**
```
    public fun consumeFuturePriceTags(
        @Payload nullableCloudEvents: List<CloudEvent?>
    ) {
```

**Stack trace and error logs**

```
Caused by: java.lang.IllegalAccessError: class io.github.stavshamir.springwolf.asyncapi.scanners.channels.annotation.SpringPayloadAnnotationTypeExtractor (in unnamed module @0x54bff557) cannot access class sun.reflect.generics.reflectiveObjects.ParameterizedTypeImpl (in module java.base) because module java.base does not export sun.reflect.generics.reflectiveObjects to unnamed module @0x54bff557
	at io.github.stavshamir.springwolf.asyncapi.scanners.channels.annotation.SpringPayloadAnnotationTypeExtractor.getPayloadParameterClass(SpringPayloadAnnotationTypeExtractor.java:34)
```

Possible solution in `SpringPayloadAnnotationTypeExtractor`:
```
        try {
            // Resolve generic type for batch listeners
            if (parameterClass == List.class) {
                Type type = ((ParameterizedTypeImpl) method.getGenericParameterTypes()[parameterPayloadIndex]).getActualTypeArguments()[0];
                return Class.forName(type.getTypeName());
            }
        } catch (Exception ex) { // Throwable here?
            log.info("Found payload type List<?>, but was unable to extract generic data type", ex);
        }
```

IllegalAccessError is not part of the caught Exception. Change catch block to Throwable?

Root cause:
`((ParameterizedTypeImpl) method.getGenericParameterTypes()[parameterPayloadIndex]).getActualTypeArguments()[0].getTypeName()` -> `? extends io.cloudevents.CloudEvent`, which is not an actual class

timonback:
@sam0r040 and I investigated and came to the conclusion that different java version are used. java8 in springwolf and java17 in the bug description.

The IllegalAccessError indicated an issue with the introduced java modules.
Add this to your java17 JVM args: `--add-opens=java.base/sun.reflect.generics.reflectiveObjects=ALL-UNNAMED"`

Related info: https://github.com/graphql-java-kickstart/graphql-java-tools/issues/265

timonback:
[...]

PR #133, opened 2023-01-20
test: GH-132: Add test for SpringPayloadAnnotationTypeExtractor

timonback:
Add log message to indicate mismatch java versions

Co-Authored-By: 93372330+sam0r040@users.noreply.github.com


stavshamir:
Thanks, maybe we should consider upgrading to java 11

PR #134, opened 2023-01-20
Add example configuration for springdoc-openapi-gradle-plugin

sam0r040:
Added an example configuration for springdoc-openapi-gradle-plugin to the springwolf-kafka-example to demonstrate how it can be used. 
To test and ensure that the plugin still works for springwolf, I also added:
* The plugin is executed before the test step
* A unit test compares the generated file with expected spec that is also used in the other test.

I'm not super happy with the testing part, because there is no way to ignore exception from springdoc-openapi-gradle-plugin. That means that if there is a problem with the application start and the plugin can not get the spec, the build will fail before the tests are executed. If that is not acceptable we can discuss more complex options ie moving this test into its own gradle task that is executed after the tests.

### DIFF
--- a/springwolf-examples/springwolf-kafka-example/src/test/java/io/github/stavshamir/springwolf/example/OpenApiGeneratorTest.java
+++ b/springwolf-examples/springwolf-kafka-example/src/test/java/io/github/stavshamir/springwolf/example/OpenApiGeneratorTest.java
@@ -1,0 +1,36 @@
+package io.github.stavshamir.springwolf.example;
+
+import org.json.JSONException;
+import org.junit.jupiter.api.Test;
+import org.skyscreamer.jsonassert.JSONAssert;
+import org.skyscreamer.jsonassert.JSONCompareMode;
+import org.testcontainers.shaded.org.apache.commons.io.IOUtils;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+
+/**
+ * This test is strongly coupled with the configuration of the springdoc openapi plugin in build.gradle
+ * <p>
+ * The gradle project is configured to run the generateOpenApiDocs task before the test task.
+ * The generateOpenApiDocs task will generate the openapi-generated.json into the test-resources so that
+ * this test can pick up the openapi-generated.json afterwards and compare it to the reference asyncapi.json
+ */
+public class OpenApiGeneratorTest {
+
+    @Test
+    public void asyncApiResourceArtifactTest() throws JSONException, IOException {
+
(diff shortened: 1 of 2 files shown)


springwolf-examples/springwolf-kafka-example/src/test/java/io/github/stavshamir/springwolf/example/OpenApiGeneratorTest.java
```diff
@@ -0,0 +1,29 @@
...
+    @Test
+    public void asyncApiResourceArtifactTest() throws JSONException, IOException {
+
+        InputStream expectedStream  = this.getClass().getResourceAsStream("/asyncapi.json");
+
+        // When running with EmbeddedKafka, localhost is used as hostname
+        String expectedWithoutServersKafkaUrlPatch = IOUtils.toString(expectedStream, StandardCharsets.UTF_8);
+        String expected = expectedWithoutServersKafkaUrlPatch.replace("kafka:29092", "localhost:29092");
+
+        InputStream actualStream = this.getClass().getResourceAsStream("/openapi-generated.json");
```
timonback, line 24: Great. Learning for me, the build folder is also scanned for resources

springwolf-examples/springwolf-kafka-example/src/test/java/io/github/stavshamir/springwolf/example/OpenApiGeneratorTest.java
```diff
@@ -0,0 +1,29 @@
...
+import org.junit.jupiter.api.Test;
+import org.skyscreamer.jsonassert.JSONAssert;
+import org.skyscreamer.jsonassert.JSONCompareMode;
+import org.testcontainers.shaded.org.apache.commons.io.IOUtils;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+
+public class OpenApiGeneratorTest {
```
timonback, line 13: Maybe put a comment that this test is linked to the build.gradle configuration and plugin?
Just to make the link more obvious

timonback (commented):
Great addition and straight forward to use

springwolf-examples/springwolf-kafka-example/src/test/java/io/github/stavshamir/springwolf/example/OpenApiGeneratorTest.java
sam0r040: I tried to put a comment to explain the connection between the gradle configuration and this test. Does it help?

timonback (approved):

PR #168, opened 2023-05-06
Export to YAML

ctasada:
Added support to create the `asyncapi` specification file in YAML.

Requesting the file to `/springwolf/docs.yaml` will provide it in YAML format. Otherwise, using the path defined by `springwolf.paths.docs` variable, concatenating `.yaml` at the end will produce the same effect


timonback:
Hi @ctasada
Thank you for this addition and keeping backwards compatability in mind!

PR #184, opened 2023-05-27
Own package per example

timonback:
Avoids conflicts (in gradle) when running the same test name for different projects


sam0r040 (approved):
Finally, thanks!

PR #191, opened 2023-05-31
Feat: Improved docket configuration

ctasada:
Looking at the AsyncAPI specification I was missing support to provide the next elements:

- [x] defaultContentType
- [x] id

The configuration for `defaultContentType` was not possible via SpringBoot Properties (only programatically)

Now we can indicate something like `springwolf.docket.default-content-type=application/json` (or `springwolf.docket.defaultContentType=application/json` to configure the value
 
It was not possible to configure the AsyncAPI identifier either (`id`)

Now we can indicate something like `springwolf.docket.id=urn:something:something` (or `springwolf.docket.id=https://something` to configure the identifier value

### DIFF
--- a/springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/types/AsyncAPI.java
+++ b/springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/types/AsyncAPI.java
@@ -4,7 +4,11 @@ import com.asyncapi.v2._0_0.model.Tag;
 import com.asyncapi.v2._0_0.model.channel.ChannelItem;
 import com.asyncapi.v2._0_0.model.info.Info;
 import com.asyncapi.v2._0_0.model.server.Server;
-import lombok.*;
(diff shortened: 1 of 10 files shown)


ctasada:
If accepted I will update https://github.com/springwolf/springwolf.github.io too

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/types/AsyncAPI.java
```diff
@@ -35,6 +35,17 @@ public class AsyncAPI {
     @Builder.Default
     private String asyncapi = "2.0.0";
```
sam0r040, line 36: Should we update this?

timonback:
Perfect. Thank you for this contribution. You are becoming a regular :)

I added a default value for `defaultContentType` and users only have to overwrite it, when it does not match.

Great that you found the documentation repository.
I am thinking of a new release in the upcoming weeks, so you get to use your contributions shortly.

PR #201, opened 2023-06-02
chore: Update AsyncAPI spec version to 2.6.0

timonback:


timonback:
Imports need to be updated as well

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/AbstractOperationDataScanner.java
```diff
@@ -1,9 +1,9 @@
 package io.github.stavshamir.springwolf.asyncapi.scanners.channels.operationdata;
 
-import com.asyncapi.v2.binding.channel.ChannelBinding;
-import com.asyncapi.v2.binding.operation.OperationBinding;
 import com.asyncapi.v2._0_0.model.channel.ChannelItem;
 import com.asyncapi.v2._0_0.model.channel.operation.Operation;
```
ctasada, line 4: I think that we should also upgrade all the `com.asyncapi.v2._0_0` imports to `com.asyncapi.v2._6_0`. Those new models are aligned with the different expectations of the new spec

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/types/AsyncAPI.java
```diff
@@ -33,7 +37,7 @@ public class AsyncAPI {
      */
     @NonNull
     @Builder.Default
-    private String asyncapi = "2.0.0";
+    private String asyncapi = "2.6.0";
```
ctasada, line 40: Non Blocking Comment: While working in another change I saw that this class is mainly a duplicated of https://github.com/asyncapi/jasyncapi/blob/master/asyncapi-core/src/main/java/com/asyncapi/v2/_0_0/model/AsyncAPI.java. Maybe would be a good idea to try to use that file directly.

ctasada (changes requested):

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/types/AsyncAPI.java
timonback: Agreed. That would be a benefit to the project and resolve some of the current issues.
If I remember correctly, the difference is the Schema type. At this point we do not have a mapping of the Jackson schema into the AsyncAPI schema.
Any help or suggestions on this are welcome.

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/AbstractOperationDataScanner.java
timonback: True, let me change the PR to draft for now.

sam0r040 (approved):

sam0r040:
@Pakisan With updating from 2.0.0 to 2.6.0 we noticed that the properties for Message and ChannelBinding have changed the value of the Bindings map from an extension of the Binding type to Object. For Example: 
```
private Map<String, ? extends MessageBinding> bindings;
```
to
```
private Map<String, Object> bindings;
```
this requires some casting for us, was this intended? Did you consider using  `Map<String,?>` instead?

Pakisan:
@sam0r040 hi!

I'll think what is the best variant to fix it. Open to your proposals, in related issue

https://github.com/asyncapi/jasyncapi/issues/126

PR #204, opened 2023-06-05
Chore/javax dependency cleanup

tvahrst:
This PR explicitly declares a `io.swagger.core.v3:swagger-core-jakarta` dependency instead of `io.swagger.core.v3:swagger-core`. 
The latter one came transiently via swagger-inflector and is now excluded.  This change allowed it to remove the dependency of `javax.xml.bind:jaxb-api`. 

swagger-inflector also pulls transiently the 'older'  `io.swagger:swagger-core` dependency (v1.6.9, other group-id, other package-names) which unfortunately cannot be excluded, because swagger-inflector needs some utility classes from this dependency (`io.swagger.util.Json`) 

This necessary old swagger-core dependency has itself a dependency to javax.validation, which should not be part of the springwolf dependency tree because springwolf changed to SpringBoot 3.x and jakarta.  So, the groupId javax.validation is explicitly excluded from the transient deps of swagger-inflector.

There was **one** usage of javax.validation.NotNull in `AsyncAPI` which seems to be accidently defined. This annotation was changed to lombok.NonNull  (like all other methods in `AsyncAPI`)

### DIFF
--- a/springwolf-examples/springwolf-cloud-stream-example/build.gradle
+++ b/springwolf-examples/springwolf-cloud-stream-example/build.gradle
@@ -44,7 +44,7 @@ dependencies {
     implementation "org.springframework.cloud:spring-cloud-stream-binder-kafka-streams"
     implementation "com.google.guava:guava:${guavaVersion}"
     implementation "org.slf4j:slf4j-api"
-    implementation "io.swagger.core.v3:swagger-core:${swaggerCoreVersion}"
+    implementation "io.swagger.core.v3:swagger-core-jakarta:${swaggerCoreVersion}"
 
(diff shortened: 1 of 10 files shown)


springwolf-examples/springwolf-cloud-stream-example/build.gradle
```diff
@@ -74,6 +74,9 @@ java {
 }
 
 test {
+    minHeapSize = "128m" // initial heap size
+    maxHeapSize = "1024m" // maximum heap size
```
tvahrst, line 78: Locally, we get an OutOfMemory during test execution in `springwolf-cloud-stream-example`. So we defined a maxHeapSize which was sufficient.

PR #206, opened 2023-06-05
feat: Remove usage of Guava

ctasada:
The Guava library is really convenient, but over time, most of its features became part
of the vanilla Java.

This PR removes the usage of Guava, replacing it by Java equivalents.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,8 +28,6 @@ ext {
 
     commonsIoVersion = '2.12.0'
 
-    guavaVersion = '32.0.0-jre'
-
     javaMoneyMonetaVersion = '1.4.2'
 
     jsonAssertVersion = '1.5.1'
--- a/springwolf-core/build.gradle
+++ b/springwolf-core/build.gradle
@@ -38,8 +38,6 @@ dependencies {
     implementation "org.springframework:spring-messaging"
     implementation "org.springframework.boot:spring-boot-autoconfigure"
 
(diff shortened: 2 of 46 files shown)


tvahrst:
Concerning MessageHelperTest: there is no garantuee, which of message2,message3 won the deduplication, so I suggest this assertion:

```
        Map<String, List<Message>> oneOfMap = (Map<String, List<Message>>) asObject;
        assertThat(oneOfMap).hasSize(1);
        List<Message> deduplicatedMessageList = oneOfMap.get("oneOf");
        // we do not have any guarantee wether message2 or message3 won the deduplication.
        assertThat(deduplicatedMessageList)
                .hasSize(2)
                .contains(message1)
                .containsAnyOf(message2, message3);
```
BTW: MessageHelper#toMessageObjectOrComposition  expects a *Set* of Messages, but Message does not implement equals/hashcode.  So the Set Sematics are not used and the Set is defacto a List of Messages. Internally, duplicates are removed via a special stream using a messageSupplier which in turn uses a TreeSet which in turn is configured with a Comparator which compares the name of the message. 
If the message name was the 'unique' id of a message, this could be easier achieved by simply implementing equals/hashcode in the Message class.

timonback (approved):
Thank you very much to go through all these occurrences and replacing them with the latest Java features in order to remove the guava dependency!

PR #207, opened 2023-06-05
replace commons.lang3.StringUtils with springframework.util.StringUtils

tvahrst:
Replaces usage of `org.apache.commons.lang3.StringUtils` with `org.springframework.util.StringUtils`.   

`org.apache.commons:commons-lang3` is not needed anymore and could be removed from dependency management (currently not declared explicityl but picked up transiently)


timonback:
Looks great, will have a look again when the PR is marked as 'ready for review'

PR #211, opened 2023-06-06
feat: Added support for Message description

ctasada:
While working at https://github.com/springwolf/springwolf-core/pull/189 we found out that the provided Message description was, in fact, the one associated to the Operation. As a result we decided to fully remove the description until we could fix it.

This PR adds support for Message description.

Any `Message` object, annotated with `@Schema` will use the annotation `description` value, as the Message description.

The https://www.springwolf.dev/docs/configuration/documenting-schemas documentation is already stating this scenario, so I don't think we need further documentation updates.


timonback:
Looks good, thanks for fixing this!

PR #214, opened 2023-06-12
chore: dependency management

harare:
This pull request takes care of the dependency management. There are two changes.

First, introduce a gradle plugin called 'ca.cutterslade.analyze'. This plugin is a dependency management tool which scans for used but undeclared dependencies and vice versa unused but declared dependencies. This results in clear and reliable dependencies in the project.

Second, add a explicit _dependencies.gradle_ file containing versions for dependencies which are used in subprojects. This approach is compatible with dependabot (see: [dependabot-gradle](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates#gradle))

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,15 @@
+buildscript {
+    // include dependency versions from explicit dependencies.gradle file
+    apply from: 'dependencies.gradle'
+}
+
+plugins {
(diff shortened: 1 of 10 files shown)


sam0r040 (approved):

sam0r040:
@harare thanks a lot for your first contribution, greatly appreciated!

PR #231, opened 2023-06-17
feat: Added @AsyncMessage

ctasada:
Added support for a new `@AsyncMessage` annotation. This annotation can be used as a method parameter annotation on the same method the `@AsyncPublisher` or `@AsyncListener` annotations are used.

This new `@AsyncMessage` annotation allows to enrich the AsyncAPI Operation Message with all the supported fields.

The Message payload documentation stays in the Swagger `@Schema` annotation. This new `@AsyncMessage` annotation is intended to enrich the missing values.

A special case is the description, which is available in both. Any value defined in `@AsyncMessage` has higher priority than any other default value.

### DIFF
--- a/springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/AbstractOperationDataScanner.java
+++ b/springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/AbstractOperationDataScanner.java
@@ -138,12 +138,12 @@ public abstract class AbstractOperationDataScanner implements ChannelsScanner {
 
             var name = annotationMessage.getName();
             if (StringUtils.hasText(name)) {
-                builder.name(annotationMessage.getName());
+                builder.name(name);
             }
 
             var title = annotationMessage.getTitle();
             if (StringUtils.hasText(title)) {
-                builder.title(annotationMessage.getTitle());
+                builder.title(title);
             }
         }
     }
--- a/springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/AbstractOperationDataScanner.java
+++ b/springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/AbstractOperationDataScanner.java
@@ -14,6 +14,7 @@ import io.github.stavshamir.springwolf.schemas.SchemasService;
 import io.swagger.v3.oas.annotations.media.Schema;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.core.annotation.Order;
+import org.springframework.util.StringUtils;
 
 import java.util.HashMap;
 import java.util.List;
@@ -100,17 +101,50 @@ public abstract class AbstractOperationDataScanner implements ChannelsScanner {
         String modelName = this.getSchemaService().register(payloadType);
         String headerModelName = this.getSchemaService().register(operationData.getHeaders());
 
-        Schema schema = payloadType.getAnnotation(io.swagger.v3.oas.annotations.media.Schema.class);
+        /*
+         * Message information can be obtained via a @AsyncMessage annotation on the method parameter, the Payload
+         * itself or via the Swagger @Schema annotation on the Payload.
+         */
+
+        var schema = payloadType.getAnnotation(Schema.class);
         String description = schema != null ? schema.description() : null;
 
-        return Message.builder()
+        var builder = Message.builder()
                 .name(payloadType.getName())
                 .title(modelName)
                 .description(description)
                 .payload(PayloadReference.fromModelName(modelName))
                 .headers(HeaderReference.fromModelName(headerModelName))
-                .bindings(operationData.getMessageBinding())
-                .build();
+                .bindings(operationData.getMessageBinding());
+
+        // Retrieve the Message information obtained from the @AsyncMessage annotation. These values have higher priority
+        //  so if we find them, we need to override the default values.
+        processAsyncMessageAnnotation(operationData.getMessage(), builder);
+
+        return builder.build();
     }
 
+    private void processAsyncMessageAnnotation(Message annotationMessage, Message.MessageBuilder builder) {
+        if (annotationMessage != null) {
+            builder.messageId(annotationMessage.getMessageId());
+
+            var schemaFormat = annotationMessage.getSchemaFormat() != null ? annotationMessage.getSchemaFormat() : Message.DEFAULT_SCHEMA_FORMAT;
+            builder.schemaFormat(schemaFormat);
+
+            var annotationMessageDescription = annotationMessage.getDescription();
+            if (StringUtils.hasText(annotationMessageDescription)) {
+                builder.description(annotationMessageDescription);
+            }
+
+            var name = annotationMessage.getName();
+            if (StringUtils.hasText(name)) {
+                builder.name(annotationMessage.getName());
+            }
+
+            var title = annotationMessage.getTitle();
+            if (StringUtils.hasText(title)) {
+                builder.title(annotationMessage.getTitle());
+            }
+        }
+    }
 }
--- a/springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/annotation/AsyncAnnotationScannerUtil.java
+++ b/springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/annotation/AsyncAnnotationScannerUtil.java
@@ -85,17 +85,36 @@ class AsyncAnnotationScannerUtil {
     public static Message processMessageFromAnnotation(Method method) {
         var messageBuilder = Message.builder();
 
-        Annotation[][] parameterAnnotations = method.getParameterAnnotations();
-        for (Annotation[] annotations : parameterAnnotations) {
-            for (Annotation annotation : annotations) {
-                if (annotation instanceof AsyncMessage asyncMessage) {
-                    messageBuilder.description(asyncMessage.description());
-                    messageBuilder.messageId(asyncMessage.messageId());
-                    messageBuilder.name(asyncMessage.name());
-                    messageBuilder.schemaFormat(asyncMessage.schemaFormat());
-                    messageBuilder.title(asyncMessage.title());
-                }
+        Annotation[] annotations = method.getAnnotations();
+        for (Annotation annotation : annotations) {
+            if (annotation instanceof AsyncListener asyncListener) {
+                return parseMessage(asyncListener.operation());
+            } else if (annotation instanceof AsyncPublisher asyncPublisher) {
+                return parseMessage(asyncPublisher.operation());
             }
+        }
+
+        return messageBuilder.build();
+    }
+
+    private static Message parseMessage(AsyncOperation asyncOperation) {
+        var messageBuilder = Message.builder();
+
+        var asyncMessage = asyncOperation.message();
+        if (StringUtils.hasText(asyncMessage.description())) {
+            messageBuilder.description(asyncMessage.description());
+        }
+        if (StringUtils.hasText(asyncMessage.messageId())) {
+            messageBuilder.messageId(asyncMessage.messageId());
+        }
+        if (StringUtils.hasText(asyncMessage.name())) {
+            messageBuilder.name(asyncMessage.name());
+        }
+        if (StringUtils.hasText(asyncMessage.schemaFormat())) {
+            messageBuilder.schemaFormat(asyncMessage.schemaFormat());
+        }
+        if (StringUtils.hasText(asyncMessage.title())) {
+            messageBuilder.title(asyncMessage.title());
         }
 
         return messageBuilder.build();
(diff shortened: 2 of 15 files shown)


springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/AbstractOperationDataScanner.java
```diff
@@ -100,17 +101,44 @@ private Message buildMessage(OperationData operationData) {
...
 
+            if (annotationMessage.getDescription() != null) {
+                builder.description(annotationMessage.getDescription());
+            }
+
+            if (annotationMessage.getName() != null) {
+                builder.name(annotationMessage.getName());
+            }
+
+            if (annotationMessage.getTitle() != null) {
```
timonback, line 137: Should we check here also for notBlank, as "" is the default value in the annotation?
Assumption: If the user does not provide something, we also do not want to set/overwrite the field

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/AbstractOperationDataScanner.java
```diff
@@ -100,17 +101,44 @@ private Message buildMessage(OperationData operationData) {
...
                 .payload(PayloadReference.fromModelName(modelName))
                 .headers(HeaderReference.fromModelName(headerModelName))
-                .bindings(operationData.getMessageBinding())
-                .build();
-    }
+                .bindings(operationData.getMessageBinding());
+
+        // Retrieve the Message information obtained from the @AsyncMessage annotation. These values have higher priority
+        //  so if we find them, we need to override the default values.
+        Message annotationMessage = operationData.getMessage();
```
timonback, line 122: This new block seems like a good opportunity to move it in an own method (i.e. processAsyncMessageAnnotation or similar)

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/AbstractOperationDataScanner.java
```diff
@@ -100,17 +101,44 @@ private Message buildMessage(OperationData operationData) {
         String modelName = this.getSchemaService().register(payloadType);
         String headerModelName = this.getSchemaService().register(operationData.getHeaders());
 
-        Schema schema = payloadType.getAnnotation(io.swagger.v3.oas.annotations.media.Schema.class);
+        /*
+         * Message information can be obtained via a @AsyncMessage annotation on the method parameter, the Payload
+         * itself or via the Swagger @Schema annotation on the Payload.
+         */
+
+        var schema = payloadType.getAnnotation(Schema.class);
```
timonback, line 109: :+1:

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/types/channel/operation/message/Message.java
```diff
@@ -20,8 +20,19 @@
...
+    private String schemaFormat = DEFAULT_SCHEMA_FORMAT;
+
+    /**
+     * Unique string used to identify the message.
+     * <p>
+     * The id MUST be unique among all messages described in the API. The messageId value is case-sensitive.
+     * Tools and libraries MAY use the messageId to uniquely identify a message, therefore, it is RECOMMENDED to
+     * follow common programming naming conventions.
+     */
+    private String messageId;
```
timonback, line 35: This field is new and can be `@Nullable` -> disregard. I guess we are not applying `@Nullable` annotations for this class.

Or do we want to provide a sensible, opinionated default (then also for all scanners) - in a different PR?

springwolf-core/src/test/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/annotation/AsyncListenerAnnotationScannerTest.java
```diff
@@ -221,6 +259,27 @@ private void methodWithMultipleAnnotation(SimpleFoo payload) {
...
+                description = "test channel operation description"
+        ))
+        private void methodWithAnnotation(
+                @AsyncMessage(
+                        description = "Message description",
+                        messageId = "simpleFoo",
+                        name = "SimpleFooPayLoad",
+                        schemaFormat = "application/schema+json;version=draft-07",
+                        title = "Message Title"
+                ) SimpleFoo payload) {
```
timonback, line 275: General question as I havent seen annotations with multiple fields being applied to method arguments much:
You added the annotation on the method argument, because it describes the payload and is independent of the operation (method)?

In springwolf, the `@Payload` annotation is the only one being used on parameters. While the other `AsyncListener` `KafkaAsyncOperationBinding`, etc are applied on the method.

If we apply it on the parameter, I wonder if it should replace the `@Payload` marker.

Using the AsyncMessage annotation on multiple parameters of the method (if there were), would result in undefined behaviour.

@sam0r040 What are your thoughts?

timonback:
I like the addition and believe it is a benefit for our users.
Let me consult with @sam0r040 for his thoughts on a couple details.

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/AbstractOperationDataScanner.java
ctasada: Makes total sense. Fixed

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/types/channel/operation/message/Message.java
ctasada: None of the fields in the class are using the `@Nullable` annotation, even when some of them are also optional. This class is also build via a Builder, so there's really nothing that should be breaking.

I'm a bit hesitant to provide a default value. The only element that we could use as really unique would be the full class name. But that seems like a poor id, since it's language related.

springwolf-core/src/test/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/annotation/AsyncListenerAnnotationScannerTest.java
ctasada: @timonback I'm happy that you ask this question, since I asked myself the same 😄 

I decided to go with a parameter annotation because, as you mention, describes the payload. But since headers are part of the `@AsyncListener` I had second thoughts about the best place.

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/AbstractOperationDataScanner.java
ctasada: Done

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/types/channel/operation/message/Message.java
timonback: Fair enough. I recently came across a code generator, which requires a `messageId` (https://github.com/asyncapi/java-spring-template)

But that is beyond the scope of this PR and even if, we can stick to the current default of null. Providing a default value could possibly be part of an (opt-in) customizer.

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/AbstractOperationDataScanner.java
```diff
@@ -100,17 +100,50 @@ private Message buildMessage(OperationData operationData) {
...
+                builder.description(annotationMessageDescription);
+            }
+
+            var name = annotationMessage.getName();
+            if (name != null && !name.isBlank()) {
+                builder.name(annotationMessage.getName());
+            }
+
+            var title = annotationMessage.getTitle();
+            if (title != null && !title.isBlank()) {
```
timonback, line 144: This can be simplified to: `StringUtils.isNoneBlank(title)`

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/annotation/AsyncAnnotationScannerUtil.java
```diff
@@ -79,4 +81,23 @@ public static Map<String, MessageBinding> processMessageBindingFromAnnotation(Me
                 .map(Optional::get)
                 .collect(Collectors.toMap(ProcessedMessageBinding::getType, ProcessedMessageBinding::getBinding));
     }
+
+    public static Message processMessageFromAnnotation(Method method) {
```
sam0r040, line 85: Can we add two tests that check this methods with and without annotation

springwolf-core/src/test/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/annotation/AsyncListenerAnnotationScannerTest.java
```diff
@@ -168,6 +171,41 @@ void scan_componentHasMultipleListenerAnnotations() {
                                 "test-channel-2", expectedChannel2));
     }
 
+    @Test
+    void scan_componentHasAsyncMethodAnnotation() {
```
sam0r040, line 175: can we add this test also to the AsyncPublisherAnnotation?

sam0r040 (changes requested):

springwolf-core/src/test/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/annotation/AsyncListenerAnnotationScannerTest.java
timonback: @sam0r040 and I checked and we think it is best to stay closer to the AsyncApi spec and provide also a similar interface to Springwolf users.
Therefore, we would add the AsyncMessage as part of the AsyncOperation annotation
(as in the spec https://www.asyncapi.com/docs/reference/specification/v2.6.0#operationObject)
(This will probably be easier to document on our website as well)

This should avoid issues like using the annotation on multiple method parameters or mixing it with unsupported annotations like KafkaListener and RabbitListener.

And, the Payload annotation can stay as is.

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/AbstractOperationDataScanner.java
ctasada: That would require adding a dependency to Apache Commons. Maybe using `StringUtils.hasText()` (from already imported Spring-Framework) would be enough.

springwolf-core/src/test/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/annotation/AsyncListenerAnnotationScannerTest.java
ctasada: Done

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/annotation/AsyncAnnotationScannerUtil.java
ctasada: Done. Also added the missing `processMessageBindingFromAnnotation` test

springwolf-core/src/test/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/annotation/AsyncListenerAnnotationScannerTest.java
ctasada: @timonback done with https://github.com/springwolf/springwolf-core/pull/231/commits/7bb542f303ec1fc027b00e3ca97868fac3f3ea18

I think that with this change all the comments are addressed

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/AbstractOperationDataScanner.java
```diff
@@ -100,17 +101,50 @@ private Message buildMessage(OperationData operationData) {
...
+            builder.schemaFormat(schemaFormat);
+
+            var annotationMessageDescription = annotationMessage.getDescription();
+            if (StringUtils.hasText(annotationMessageDescription)) {
+                builder.description(annotationMessageDescription);
+            }
+
+            var name = annotationMessage.getName();
+            if (StringUtils.hasText(name)) {
+                builder.name(annotationMessage.getName());
```
timonback, line 141:
```suggestion
                builder.name(name);
```

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/AbstractOperationDataScanner.java
```diff
@@ -100,17 +101,50 @@ private Message buildMessage(OperationData operationData) {
...
+            }
+
+            var title = annotationMessage.getTitle();
+            if (StringUtils.hasText(title)) {
+                builder.title(annotationMessage.getTitle());
```
timonback, line 146:
```suggestion
                builder.title(title);
```

timonback (approved):

sam0r040 (approved):

timonback:
Great addition and thank you for the review cycles. Looking forward to release this soon and supporting more AsyncApi features

PR #242, opened 2023-06-26
build: Fixed Swagger-Core dependencies

ctasada:
https://github.com/swagger-api/swagger-core is a monorepo and all the group dependencies are released together.

With the merge of https://github.com/springwolf/springwolf-core/pull/240 a breaking change was introduced and the code was not compiling anymore. Also, for some reason DependaBot was not identifying the need to update the rest of the dependencies.

Now we align all the Swagger-Core related dependencies.

PR #243, opened 2023-06-26
GH-242: chore: Reduce jackson dependency to single one

timonback:
com.fasterxml.jackson is also released together, so each artifact has the same version


timonback:
Relates to https://github.com/springwolf/springwolf-core/pull/260

PR #245, opened 2023-06-26
docs: Describe how to use snapshot builds with maven

timonback:


ctasada:
@timonback I love the idea to document it and release SNAPSHOT versions, but I think we should adopt the standard version lifecycle.

`0.1-SNAPSHOT -> 0.1 -> 0.2-SNAPSHOT`

Right now we have, for example `0.11.0` released, while we continue to publish `0.11.0-SNAPSHOT` versions, which is confusing.

timonback:
Agreed @ctasada, we comply to the standard. I will move it #258 as I like to simplify the version strategy as well.

PR #264, opened 2023-07-01
support channel server assignments

tvahrst:
With this PR, `@AsyncOperation` was extended allowing to optionally define server name(s) for a channel. This is necessary if the asyncapi defines multiple servers. 

Example
annotated method
```
        @AsyncListener(operation = @AsyncOperation(
                channelName = "channel1"",
                description = "description",
                servers = {"kafka"}
        ))
        private void methodWithAnnotation(String payload) {
        }

```
resulting asyncapi
```
  ...
  "channels": {
    "channel1": {
      "description": "description",
      "servers": [
        "kafka"
      ],

```


springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/annotation/AsyncPublisherAnnotationScanner.java
```diff
@@ -75,16 +75,17 @@ private Stream<OperationData> toOperationData(Method method) {
         Class<AsyncPublisher> annotationClass = AsyncPublisher.class;
         return Arrays
                 .stream(method.getAnnotationsByType(annotationClass))
-                .map(annotation -> toConsumerData(method, operationBindings, messageBindings, annotation));
+                .map(annotation -> toProducerData(method, operationBindings, messageBindings, annotation));
     }
 
-    private ProducerData toConsumerData(Method method, Map<String, OperationBinding> operationBindings, Map<String, MessageBinding> messageBindings, AsyncPublisher annotation) {
+    private ProducerData toProducerData(Method method, Map<String, OperationBinding> operationBindings, Map<String, MessageBinding> messageBindings, AsyncPublisher annotation) {
```
tvahrst, line 81: the method returns ProducerData, corrected the method name (there is a identical method for ConsumerData so I assume a copy/paste effect ;-))

tvahrst:
oh, it seems that my IDE has different settings for import-handling. Is there a recommendation for when to use *?

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/annotation/AsyncOperation.java
```diff
@@ -25,6 +25,14 @@
      */
     String description() default "";
 
+    /**
+     * The servers on which this channel is available, list of names of Server Objects.
+     * If servers is empty then this channel is available on all servers defined in the Async API.
+     * Mapped to (
```
timonback, line 31: Looks like some text got missing

timonback:
Great addition.

Regarding the wildcard import, at this point it is up to us. But I like to introduce formatting rules so that this is automatically handled -> https://github.com/springwolf/springwolf-core/pull/263

I am wondering if we want to add validation, so that any server defined in the annotation must be part of the servers definition (accessible through `AsyncApiDocketService`)
In the end, we promise to generate compliant AsyncApi documents.

tvahrst:
> Great addition.
> 
> Regarding the wildcard import, at this point it is up to us. But I like to introduce formatting rules so that this is automatically handled -> #263
> 
> I am wondering if we want to add validation, so that any server defined in the annotation must be part of the servers definition (accessible through `AsyncApiDocketService`) In the end, we promise to generate compliant AsyncApi documents.

tvahrst:
ui, sorry, unintentionally closed the PR.

tvahrst:
Adding validation of server entries is a good idea - I will try to add this to the PR. BTW: what should happen if a server-name on a channel does not match the list of definied servers? Log a warning, throw an exception, delete the server-mapping or add a hint to the server-mapping?

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/annotation/AsyncOperation.java
tvahrst: ups, I take a look...

tvahrst:
Currently, the complete scanning and detection of channels happens during spring-bean creation of  `DefaultChannelService` (in the `@PostConstruct` annotated method `findChannels`. 

At this point, we do not have any access to other beans or informations of other beans. So it is not possible to validate channel values (i.e. the server mappings) during channel detection/scanning because we don't have access to the `Docket` instance holding the list of defined servers.

The only location to implement a validation is in DefaultAsyncApiService in ` buildAsyncAPI` after the final AsyncApi builder.call. So I suggest to implement a kind of 'postvalidation' at this location.

timonback:
> Adding validation of server entries is a good idea - I will try to add this to the PR. BTW: what should happen if a server-name on a channel does not match the list of definied servers? Log a warning, throw an exception, delete the server-mapping or add a hint to the server-mapping?
> 
> 

In the past when we found an invalid configuration, we throw an exception. The server name is added by the user explicitly, so we can expect to configure it correctly. Might even be best to get feedback that the configuration does not work.

timonback:
Hi @tvahrst,
we are slowing coming back from vacation and starting to plan the next release.
Is there any kind of support you need from us at this point?

tvahrst:
There have been quite some changes since july ;-) I will update (merge master) the pr-branch and take a look to provide the earlier discussed validation of server-names.

tvahrst:
Due to previous changes with spring bean initialization (#292) it was quite simple to add a validation of server refs provided on operations/channels with the servers defined in the AsyncAPI `servers` block.

The main validation happens in `AbstractOperationDataScanner`. 

During implementation of the validation I noticed two things:

1. The `DefaultAsyncApiDocketService` has a service method `getAsyncApiDocket()` which provides the current Docket instance. If the application defines a Docket bean, `DefaultAsyncApiDocketService` holds a reference on this bean and delivers the bean on every `getAsyncApiDocket()` invocation.   If there is no Docket bean, `DefaultAsyncApiDocketService` creates an `AsyncApiDocket` object with properties from environment. This happens again with every invocation of `getAsyncApiDocket()`.  As there are many 'uses' of this service, I would to suggest to cache a environment-based Docket once created. 

2. `AbstractOperationDataScanner` is an abstract superclass for different concrete sub-classes which implements different strategies to obtain OperationData. This inheritance pattern makes it difficult to write unit-tests for the methods in the abstract class (in this case the `validateServers()` method. Additionally, the access to references to other spring beans (`SchemasService`, `AsyncApiDocketService`)  is designed via abstract getter-methods which must be implemented in every subclass. I would like to provide a proposal for an alternative approach, replacing inheritance with a composition pattern. I will put this in an seperatore PR for discussion...

timonback (approved):

PR #284, opened 2023-07-11
chore: Reduce duplicities in configuration

ctasada:
With the recent changes in https://github.com/springwolf/springwolf-core/pull/267 now is possible to reduce duplication of values.

Now the version is reduced to a single `build.gradle` configuration. Same for other module configurations that are also unified.

While testing I also found some unneeded variables in `dependencies.gradle` and some `unusedDeclaredArtifacts` which are also fixed.

Each change is in its own commit to allow for easier review.


.env
```diff
@@ -0,0 +1 @@
+SPRINGWOLF_VERSION=0.13.0
```
ctasada, line 1: This file could directly be `0.13.0-SNAPSHOT` and remove the need for the `-Dsnapshot=true` flag, but that requires some further changes in the release process.

The "standard" Maven release process is usually:
- Remove the `SNAPSHOT` suffix
- Release
- Increase version + add the `SNAPSHOT` suffix

I leave this change open for discussion

build.gradle
```diff
@@ -36,9 +46,95 @@ allprojects {
...
 }
 
 dependencyCheck {
     outputDirectory = "$buildDir/reports/owasp-dependency-check"
     formats = [ReportGenerator.Format.SARIF.toString()]
-}
\ No newline at end of file
+}
+
+def static getenv(path = ".env") {
```
timonback, line 125: Can we rename this to some like `readEnvFile` or similar?
getenv seems to be related to a single env variable: https://man7.org/linux/man-pages/man3/getenv.3.html

For everyone not so familiar with gradle, it might be good to define a function for the line above as well:
`getenv().each { key, value -> if (System.getProperty(key as String) == null) { System.setProperty(key as String, value as String) } }`
Maybe `applyDotEnvFileAsGradleProperties()`?

timonback:
Hi @ctasada,
I like the change to reduce the amount of version string sprinkled around in this repo.

I like to keep the information whether it is a snapshot separate from `SPRINGWOLF_VERSION`, because I can be sure in the pipeline that builds on `master` will never build release version.

What I am currently unsure about is how to handle bugfix releases.
Assuming we release 0.13.0 and incremented to the next version 0.14.0. Now, we discover something requiring a bugfix aka small release 0.13.1. What to do?
Fix it in 0.14.0, cherry-pick backport it to a new branch created from the 0.13.0 tag (with increasing the version to 0.13.1 on the newly created branch first)?

ctasada:
> 

@timonback Regarding the bug fixing after releases I think the best approach is to link the release with a `tag`.

I see 2 ways of accomplishing it:
1.- Modify the Github workflow in a way that once the release is generated, it also creates a git `tag` This will allow to checkout and build a new tag with any needed bugfix in the future
2.- Modify the Github workflow so the release process is directly related to the tag creation. So once a new tag is created, the libraries release is executed and the version is updated.

Both approaches accomplish the same outcome, but are implemented in slightly different ways. I see that Github has different workflows in the marketplace that accomplish this. I have used both approaches with Gitlab for years and they work fine. The decision is mainly driven by minor tradeoffs and deployment preferences.

build.gradle
ctasada: @timonback modified the code with the proposed approach

.env
timonback: On second thoughts, I see the benefit of it and like suggestion.
We will merge this, feel free to suggest a change, to further simplify this.

I enjoy all the thoughts and ideas you bring to this project.

timonback: Based on this issue https://github.com/springwolf/springwolf-core/issues/295 it will reduce confusion which version is currently used.
So, one more argument for your suggestion :)

Issue #288, opened 2023-07-14
Provide support to configure the `bindings` in the `@AsyncOperation`


ctasada:
**Describe the feature request**
SpringWolf provides plugins for Kafka or RabbitMQ bindings, but there's no other way to configure binding via the annotations.

Having something like
```java
@AsyncOperations(
    bindings = {@GooglePubSubBinding(
        topic = "my-topic",
    )}
)
```

We may need to add different annotations to represent the different bindings, but there's no need to add all the bindings at the same time.

**Motivation**
The provided plugins are really convenient, but when an unsupported binding is needed, the current approach is limiting.

**Technical details**
I'm happy to provide a PR providing a first batch of binding annotations. For sure we need to consider cases where the plugins may be "conflicting" in which case we need to define which definitions takes precedence. I think the information in the annotation should probably overwrite any other autogenerated information.

**Describe alternatives you've considered**
Providing plugins for each binding seems overkill, mainly when some of those of those bindings may not have specific Spring supported annotations, like GooglePubSub.

timonback:
@sam0r040 and me came across the issue some time ago.

Due to limitations of combining annotations and generics (`bindings`=@GooglePubSubBinding) we opted for an additional annotation.
Our solution was the addition of kafka and amqp specific OperationBinding operations. See https://www.springwolf.dev/docs/configuration/documenting-consumers#amqpasyncoperationbinding

For sure what is missing currently is a generic annotation to let users simply define the (basic) binding.

Brandom2725:
It's an issue with some wallets apparently or proof of your lost digital assets I had a similar issue. I contacted MATHEW LEWIS DEVELOPER directly and they had it resolved for me. 
mathewlouisdeveloper@gmail.com
Send them a mail or Contact them on Twitter @mathew_dev1 They'll help you out!

timonback:
Thank you for the report, the issue has been addressed in the new release.

Docs will follow soon on the website, proposed documentation: https://github.com/springwolf/springwolf.github.io/pull/62/files#diff-07b414902f2f07f5c083d5a81a4ee3730ad48b61787224581b2160a2f5b13c05

Feel free to reopen this issue if there is still something missing.

PR #296, opened 2023-07-20
feat: Added an `@AsyncOperationBinding` annotation to simplify the creation of new operation bindings

ctasada:
So far we had 2 Operation Binding implementations: `springwolf-kafka-plugin` and `springwolf-amqp-plugin`. Now we have a new annotation that can be used to simplify the implementation of new bindings, reducing code duplicities.

Any new `OperationBindingProcessor` can extend from the new `AbstractOperationBindingProcessor`.

This PR is created as a first draft to follow up from the https://github.com/springwolf/springwolf-core/issues/288#issuecomment-1636082284 comment.

@timonback @sam0r040 What do you think of this approach? Would make sense to do something similar with the `MessageBindingProcessor`? My next step would be to create a new `GooglePubSubBinding` implementation, probably as its own plugin, using this approach.

### DIFF
--- a/springwolf-core/build.gradle
+++ b/springwolf-core/build.gradle
@@ -55,6 +55,7 @@ dependencies {
     testCompileOnly "org.projectlombok:lombok:${lombokVersion}"
 
     testImplementation "org.junit.jupiter:junit-jupiter-api:${junitJupiterVersion}"
+    testImplementation("org.junit.jupiter:junit-jupiter-params:${junitJupiterVersion}")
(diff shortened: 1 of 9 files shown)


sam0r040 (changes requested):
We really like the change and the prospect of adding more generic bindings. However, it feels like there are test cases missing for OperationBinding. It would be really great if you could add some in AsyncAnnotationScannerUtilTest.

ctasada:
@sam0r040 I'm back from vacation and picking up again this issue. Which test cases do you think are missing?

sam0r040 (approved):

Issue #295, opened 2023-07-20
Cloud not run docker compose up on spring-cloud-stream-example


nguyenminhtuanfit:
Error response from daemon: manifest for stavshamir/springwolf-cloud-stream-example:0.13.0 not found: manifest unknown: manifest unknown

sam0r040:
Hi, we have not published version 0.13.0 of the cloud stream example yet. You can either use the snapshot or an older version. You can see the published versions here: https://hub.docker.com/r/stavshamir/springwolf-cloud-stream-example/tags

timonback:
Thank your for the report, 0.13.0 has been released and we have improved the project to avoid confusions like these in the future.

Feel free to reopen this issue if there is still something missing.

PR #299, opened 2023-07-23
Chore/handling of snapshot builds

timonback:
Thank you for the suggestion @ctasada, it indeed makes things simpler and moves us closer to the standard workflow when working with snapshots


timonback:
Relates to: https://github.com/springwolf/springwolf-core/issues/295

Issue #176, opened 2023-05-22
KafkaMessageBinding annotation does not use bindingVersion default value


timonback:
**Describe the bug**
When using the KafkaMessageBinding annotation without any parameters, the builders default value for bindingVersion gets set to null - although the default version should be used

Example AsyncAPI document should have an operation.BindingVersion of 0.4.0 as defined in KafkaMessageBinding#bindingVersion

**Dependencies and versions used**
For example, `springwolf-kafka` version `0.11.0`.

**Code example**
```
    @AsyncPublisher(operation = @AsyncOperation(
            channelName = topicName,
    ))
    @KafkaAsyncOperationBinding // <- no parameter set
    public void sendMessage(AnotherPayloadDto msg) {
        kafkaTemplate.send(topicName, msg);
    }
```

**Stack trace and error logs**
None

Related code: `KafkaMessageBindingProcessor#mapToMessageBinding`

PR #300, opened 2023-07-23
Fix/gh 176 keep default kafka message binding when null

timonback:


timonback:
Closes https://github.com/springwolf/springwolf-core/issues/176

PR #316, opened 2023-08-07
GH-134: fix: Handle NPE in ExampleJsonGenerator

timonback:
i.e. A MapSchema is an object, but doesn't have properties


timonback:
Although not supported by AsyncApi natively, should we add support for the missing swagger schema types?

```
// bytearray
// composed
// email
// file
// json
// map
// uuid
```

PR #370, opened 2023-09-25
feat(core): add AsyncGenericBinding

timonback:
Relates to https://github.com/springwolf/springwolf-core/issues/288, but in a generic, manual way


springwolf-examples/springwolf-sqs-example/src/main/java/io/github/stavshamir/springwolf/example/sqs/producers/AnotherProducer.java
```diff
@@ -21,6 +22,9 @@ public class AnotherProducer {
                             channelName = QUEUE,
                             description = "Custom, optional description defined in the AsyncPublisher annotation"))
     @SqsAsyncOperationBinding
+    @AsyncGenericOperationBinding(
+            type = "custom-sqs",
+            fields = {"internal-field=customValue"})
```
sam0r040, line 27: Reminder: Add nested field

PR #381, opened 2023-09-29
Refactor/avoid user controlled input

timonback:


timonback:
The schemaService does not return the FQN of the java class, therefore this code does not work as expected as this point.

Idea for the next release: We change the schema name to the FQN of the class.
This also resolves the issue when multiple classes with different FQN, but same simpleName are found that both exist in the final AsyncApi document (instead of last-write-wins)

springwolf-plugins/springwolf-kafka-plugin/src/main/java/io/github/stavshamir/springwolf/asyncapi/controller/SpringwolfKafkaController.java
```diff
@@ -29,7 +30,7 @@
 @ConditionalOnProperty(prefix = SPRINGWOLF_KAFKA_CONFIG_PREFIX, name = SPRINGWOLF_KAFKA_PLUGIN_PUBLISHING_ENABLED)
 public class SpringwolfKafkaController implements InitializingBean {
 
-    private final AsyncApiDocketService asyncApiDocketService;
+    private final DefaultSchemasService schemasService;
```
timonback, line 33: Use SchemasService interface

springwolf-plugins/springwolf-amqp-plugin/src/main/java/io/github/stavshamir/springwolf/asyncapi/controller/SpringwolfAmqpController.java
```diff
@@ -29,7 +30,7 @@
 @ConditionalOnProperty(prefix = SPRINGWOLF_AMQP_CONFIG_PREFIX, name = SPRINGWOLF_AMQP_PLUGIN_PUBLISHING_ENABLED)
 public class SpringwolfAmqpController implements InitializingBean {
```
timonback, line 31: Extract base conversion methods to core to reduce duplication?

timonback: Will do in the future

PR #393, opened 2023-10-06
feat(core): add configuration option to use fully qualified names in schemas

sam0r040:
The  default is false (unchanged behavior)

### DIFF
--- a/springwolf-core/src/main/java/io/github/stavshamir/springwolf/schemas/DefaultSchemasService.java
+++ b/springwolf-core/src/main/java/io/github/stavshamir/springwolf/schemas/DefaultSchemasService.java
@@ -17,12 +17,14 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Function;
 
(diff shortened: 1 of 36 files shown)


springwolf-core/src/main/java/io/github/stavshamir/springwolf/schemas/DefaultSchemasService.java
```diff
@@ -25,10 +27,17 @@ public class DefaultSchemasService implements SchemasService {
...
 
-    public DefaultSchemasService(List<ModelConverter> externalModelConverters, ExampleGenerator exampleGenerator) {
+    public DefaultSchemasService(
+            List<ModelConverter> externalModelConverters,
+            ExampleGenerator exampleGenerator,
+            SpringwolfConfigProperties properties) {
 
         externalModelConverters.forEach(converter::addConverter);
         this.exampleGenerator = exampleGenerator;
+
```
timonback, line 37: After the swagger core update, getfqn can be called to fetch the current flag status which allows the reset to the original value

Issue #395, opened 2023-10-13
Allow info-extensions in application.yml/props


tvahrst:
Springwolf supports the definition of  'static' basic informations of an asyncapi spec (info, id, servers) by
* providing an Docket Bean  or
* providing the necessary informations as environment properties (for example in application.yml) with the prefix `springwolf.docket.*`. 

The second approach (docket data via environment)  does currently not support the definition of extensions for the info-block. So it isn't possible to provide docket data so that the resulting asyncapi looks like this (info include `x-additional-field`):

```
{
  "asyncapi": "2.6.0",
  "info": {
    "title": "my-kafka-worker",
    "version": "1.0.0",
    "x-additional-field": "trulallal"
  },
  ...
}
```

For `contact` and `license`  in turn there is a possibility to define extensions in the environmet, like this:

```
springwolf.docket.info.contact.extension-fields.x-userid = t.vahrst
```

I suggest to support the same extension pattern which allows extensions for 'contact' and 'license' to be appliable for 'info'.

timonback:
Thank you for the report, the issue has been addressed in the new release.

An update on the documentation is welcome: https://github.com/springwolf/springwolf.github.io/issues/58

Feel free to reopen this issue if there is still something missing.

PR #401, opened 2023-10-13
#395 support info-extensions definition in environment

tvahrst:
This PR contains a little enhancement which allows the definition of extensions (x-...) for the asyncapi-info block in the spring environment (i.e. in application.yml / .properties)


springwolf-core/src/main/java/io/github/stavshamir/springwolf/configuration/DefaultAsyncApiDocketService.java
```diff
@@ -64,19 +64,25 @@ private AsyncApiDocket parseApplicationConfigProperties(SpringwolfConfigProperti
         return builder.build();
     }
 
-    private static Info buildInfo(@Nullable SpringwolfConfigProperties.ConfigDocket.Info info) {
```
tvahrst, line 67: This method has an input variable `info` and now a new local info variable as return type. I renamed the variables to make clear, which one is the configdocketinfo and which one is the resulting asyncapiInfo Object.

timonback:
Thanks @tvahrst for the addition of the extension object in the info docket.

We have two questions:
1. We see that the extension property is quite generic (`Map<String, Object>`) and are wondering which use-case you are covering with it. We can imagine a simple `Map<String, String>` pair, for which we can simplify the type. Alternatively, the nested structure that you used in the PR. If that is required, we would like to see also a test case along with it.
2. At this point, the extensions are part of the docket. We were looking for changes in the AsyncApi document as well, but haven't found code changes for that. Feel free to adapt one of the example projects, so that the property is also present in the `asyncapi.json` file.

springwolf-core/src/main/java/io/github/stavshamir/springwolf/configuration/DefaultAsyncApiDocketService.java
```diff
@@ -58,19 +58,25 @@ private AsyncApiDocket parseApplicationConfigProperties(SpringwolfConfigProperti
...
-                .license(info.getLicense())
+        Info asyncapiInfo = Info.builder()
+                .version(configDocketInfo.getVersion())
+                .title(configDocketInfo.getTitle())
+                .description(configDocketInfo.getDescription())
+                .contact(configDocketInfo.getContact())
+                .license(configDocketInfo.getLicense())
                 .build();
+
+        asyncapiInfo.setExtensionFields(configDocketInfo.getExtensionFields());
```
tvahrst, line 78: this is the location where the extensions are moved from docket into asyncapi

springwolf-core/src/test/java/io/github/stavshamir/springwolf/configuration/DefaultAsyncApiDocketServiceIntegrationTest.java
```diff
@@ -32,6 +32,7 @@ public class DefaultAsyncApiDocketServiceIntegrationTest {
                 "springwolf.enabled=true",
                 "springwolf.docket.info.title=Info title was loaded from spring properties",
                 "springwolf.docket.info.version=1.0.0",
+                "springwolf.docket.info.extension-fields.x-api-name=api-name",
```
tvahrst, line 35: Test of info extension

tvahrst:
Hi Timon,
I was inspired by the extension-Type `Map<String, Object` which is used in the asyncapi classes (Info, Contact etc.). But I agree, changing this to `Map<String, String>` in the docketconfig-class is the better solution as `SpringWolfConfigProperties#ConfigDocket` takes all values from the spring environment. I will change this.

The mapping between ConfigDocket#Info  and AsyncAPI happens in `DefaultAsyncApiDocketService`. I added on line to map the extensions (see comment above).

And I extended one of the existing Tests to check, whether an extension defined in the environment will be present in the resulting asyncapi (see comment above)

springwolf-core/src/test/java/io/github/stavshamir/springwolf/configuration/DefaultAsyncApiDocketServiceIntegrationTest.java
timonback: 👍 I would like to see that these fields are also serialized into the json document.
We use the example projects to demonstrate, test and verify features. Can you add a field to an example (i. e. amqp) and update the Asyncapi.json?

sam0r040: I think there was a misunderstanding and you did much more than we would have expected. We do not need to test the manually defined docket vs the docket build from spring properties again in an example project. This should be/is already done in core. I think what Timon actually meant: We can just configure extension fields in the application properties (like you already did) and just add those to the original asyncapi.json file :). Sorry for the confusion but I think it makes your change much simpler...

sam0r040:
@tvahrst For me this is ready for merging if you reduced the testing in the amqp example (see my other comment).

springwolf-core/src/test/java/io/github/stavshamir/springwolf/configuration/DefaultAsyncApiDocketServiceIntegrationTest.java
tvahrst: I am not quite sure what you mean :slightly_smiling_face: . Do you mean to replace the custom springwolf `Info` class in `SpringWolfConfigProperties` with the `com.asyncapi...Info` class, so that properties from application.yml are mapped directly to the 'info' part of the asyncapi?  

If this is the case: I wondered already why `SpringWolfConfigProperties` uses  `com.asyncapi...Contact` and `com.asyncapi...License` while `Info` is implemented as custom springwolf class.  I assumed, the different 'Nullable' contracts of `title` and `version` were the reason for this design decision. (?)

If I replace springwolf `Info` with `com.asyncapi...Info` there will be a slightly different contract, title and version have to be provided than in application.yml.

Please give me a quick signal if this is what you mean and whether I can simply the SpringWolfConfigPropertis Info part...

sam0r040: No that's not what I meant😅. I will try to write it down more clearly tomorrow...

sam0r040: Hi @tvahrst, so my original comment was about the number of test cases. We would only have expected the change you did in AsyncApiConfiguration, so that users can see how the spec would look like with the extension fields. I had another look at the code and especially the setup of the example, and I think I can now understand why you added the second test case where the extensions would be set in application.properties. As it is just an example project for demonstration purposes, we do not want to do extensive testing of core functionality there to keep maintenance efforts low.

I can see the point of adding the example where the extensions are set by spring properties instead of static configuration in the code, because thats the feature you wanted to add. I would suggest to keep the PR as it is for now, I will discuss with Timon how we want to deal with this cases and we will then either merge the PR as is or remove one of test cases. Is that okay for you?

tvahrst: Of course, thats ok. 👍

timonback (approved):

sam0r040:
@tvahrst Superseeded by https://github.com/springwolf/springwolf-core/pull/423

Please enable so that we can do rebasing on your PRs in the future:
Screenshot 2023-10-27 at 15 34 18

timonback:
@tvahrst In case you find time, feel free to write a couple sentences for the documentation -> https://github.com/springwolf/springwolf.github.io/issues/58

PR #412, opened 2023-10-18
example of changing inheritance to composition

tvahrst:
This PR is an example for an refactoring of an abstract superclass (here `AbstractOperationDataScanner`) to an 'Utils-Bean' which is uses as dependency in the concrete implementation classes.  It is mainly intended as a basis of discussion.


sam0r040:
Hi @tvahrst, 
we had a longer discussion about this. We like the change towards composition. 

This opens up more extension points to intercept into the api, we still treat the bean interfaces as internals. Consumers would be able to alter behavior, but we currently do not plan to actively support it as a public api.  

We were wondering about your motivation for this change, is it just a refactoring to introduce more best practices or do you have a specific need? 

We know this is a draft, but we could imagine to split up OperationDataScannerUtil even more to separate building channels and messages. First guess would be something like ChannelItemsUtil, ChannelItemBuilder, MessageBuilder, but there are probably better alternatives.

tvahrst:
HI, a few days ago I made some enhancements to an old pr #264 and I experienced some issues during implementation which I commend there: (https://github.com/springwolf/springwolf-core/pull/264#issuecomment-1766333788). This was the main motivation for this little pr.

I just saw (#422) that AsyncApiDocket and some DataScanner classes are deprecated. So the initial reason for this pr (`AbstractOperationDataScanner`) will be removed in the future.

timonback:
Closing as a lot of code has changed by now.
Thank you for the inspiration of composition as we used it as a base to do refactorings to all scanners!

PR #422, opened 2023-10-27
chore(core): deprecated AsyncApiDocket bean

timonback:
Use spring properties instead

### DIFF
--- a/springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/AbstractOperationDataScanner.java
+++ b/springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/AbstractOperationDataScanner.java
@@ -26,6 +26,7 @@ import static java.util.stream.Collectors.toMap;
 import static java.util.stream.Collectors.toSet;
 
 @Slf4j
+@Deprecated(forRemoval = true)
 public abstract class AbstractOperationDataScanner implements ChannelsScanner {
 
     protected abstract SchemasService getSchemaService();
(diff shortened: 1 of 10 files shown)


maksymgendin:
@timonback Couldn't find a better place for commenting on this.

We're currently using manual configuration via `AsyncApiDocket` and I'm not sure what should be an alternative for our use case because we don't use the default Spring Kafka integration where method signature and annotations are used to configure the Kafka consumers, producers and payloads. We're have a fully reactive application with `io.projectreactor.kafka:reactor-kafka` where the entry point in consuming Kafka records is a `ReactiveKafkaConsumerTemplate`:

```
private <T> ReactiveKafkaConsumerTemplate<String, T> buildReactiveKafkaConsumerTemplate(Class<T> eventClass,
                                                                                        KafkaProperties kafkaProperties,
                                                                                        String topicName,
                                                                                        MeterRegistry meterRegistry) {
    return new ReactiveKafkaConsumerTemplate<>(
            ReceiverOptions.<String, T>create(kafkaProperties.buildConsumerProperties())
                    .withKeyDeserializer(new ErrorHandlingDeserializer<>(new StringDeserializer()))
                    .withValueDeserializer(new ErrorHandlingDeserializer<>(new JsonDeserializer<>(eventClass, false)))
                    .consumerListener(new MicrometerConsumerListener(meterRegistry))
                    .subscription(Collections.singleton(topicName))
    );
}
```

`ReceiverOptions` is a class from `reactor-kafka` as well.

Could you please give me any hints how Springwolf is supposed to work with configuration only in our case?

maksymgendin:
One more problem is that we're using Springwolf mainly for local development with Testcontainers and the `servers` property is currently set dynamically on startup when constructing the `AsyncApiDocket` bean. How this can work with Spring properties?

timonback:
Hi @maksymgendin,
feel free to comment in #445 in case you run into issue. We improved our migration guide.

PR #430, opened 2023-11-03
feat(core): add library name to info object extension field

timonback:
Depends on https://github.com/springwolf/springwolf-core/pull/466


timonback:
Todo: check why cs plugin does not abort after some time

PR #434, opened 2023-11-05
fix: toClassObject method to return correct type in case of parameterised type

SheheryarAamir:
Fix an issue,  where I was expecting MyEvent as payload type but instead I am getting org.springframework.messaging.Message as payload type which is wrong

below is the signature of my consumer

@Bean
    public Consumer<Message<MyEvent>> consumeEvent() {
    }


stavshamir:
Hi @SheheryarAamir, thanks for creating this PR!
Reading the issue description, it seems like your problem is specifically with the Generic payload type when using Spring's `Message<T>` class. 

I believe the use of this class is a special case, and we would probably not want to address every use of generic payload like we would address the use of `Message<T>`.

I added some comments in the code that may help you improve this PR with the aim of solving your issue.
Thanks and please let us know if you have any other question or need any help.

springwolf-plugins/springwolf-cloud-stream-plugin/src/test/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/cloudstream/FunctionalChannelBeanDataTest.java
```diff
@@ -89,7 +89,8 @@ void testConsumerBeanWithGenericPayload() throws NoSuchMethodException {
         Set<FunctionalChannelBeanData> data = FunctionalChannelBeanData.fromMethodBean(method);
 
         assertThat(data)
```
stavshamir, line 91: This is not a batch listener - it is intended that the entire payload here is a List.
Please create a new bean method where the Generic type of the `Consumer` is `Message<String>` and add a specific test for this case.

springwolf-plugins/springwolf-cloud-stream-plugin/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/cloudstream/FunctionalChannelBeanData.java
```diff
@@ -79,8 +79,16 @@ private static Class<?> toClassObject(Type type) {
             if ("org.apache.kafka.streams.kstream.KStream".equals(rawType.getTypeName())) {
                 return (Class<?>) ((ParameterizedType) type).getActualTypeArguments()[1];
             }
+            // Resolve generic type for batch listeners
+            if (rawType == List.class) {
```
stavshamir, line 83: This will not solve your issue - but it is close. Instead of checking for raw type  equal to `List`, you should check equality to `Message`. 

You could also improve the readability and maintainability of this code by extracting the handling of this case to another method.

Issue #433, opened 2023-11-05
Spring Cloud Stream consumer with parameterised type not returning the correct payload type


SheheryarAamir:
**Describe the bug**
I am facing another issue, where I am expecting MyEvent as payload type but instead I am getting org.springframework.messaging.Message as payload type which is wrong

below is the signature of my consumer 
```
@Bean
    public Consumer<Message<MyEvent>> consumeEvent() {
    }
 ```

**Dependencies and versions used**
`springwolf-cloud-stream` version `0.6.1`.

**Code example**
```
@Bean
    public Consumer<Message<MyEvent>> consumeEvent() {
    }
 ```

**Stack trace and error logs**
N/A

timonback:
Hi @SheheryarAamir,
as mentioned above, we are not sure that we found the best solution in #441.

Still, we wanted to include this in Springwolf 0.17.0, therefore the current solution is included and marked as beta.
More details soon on the website or in https://github.com/springwolf/springwolf.github.io/pull/65

To avoid changing behaviour, we kept `List` as part of the defaults.

Thank you for opening the issue and starting on the PR.
We would love to hear from you if this resolves your issue and what is not optimal yet so that we can improve.

SheheryarAamir:
Thank you @timonback. Yes, solution implemented in version `0.17.0` partly resolve my issue as we are able to extract payload type in case it is wrapped in spring `Message` interface. But current version does not support custom interfaces e.g 
`MyInterface<String>`

timonback:
Great.
Have you tried the new configuration documented at https://www.springwolf.dev/docs/configuration/documenting-messages#unwrapping-the-payload ?
Instead of the java util Function, you will have to use your own fully qualified class name.

SheheryarAamir:
Thank you @timonback. Its working, able to extract payload type from custom interface.

PR #441, opened 2023-11-08
feat: interceptor filter pattern for serialization/deserialization common cases such as Message and KStream 

SheheryarAamir:
Implement the interceptor filter pattern. Each type has its own class (message, kstream, ...). This allows users to specify their own or (de)activate them. And it opens up the extension point via bean injection and ordering

Relates to https://github.com/springwolf/springwolf-core/issues/433


timonback:
Hi @SheheryarAamir 
@sam0r040 and I started on it by converting the SpringPayloadExtractor in core to a spring bean.

The idea is that List-Extractor, Message- (and KStream- in cloud-stream) would implement the newly introduced interface.
The SpringPayloadExtract finds them via dependency injection (List<PayloadExtractor>).

This might be a bit more tricky than expected.
Next steps would be:
- Create a class for the KStream- & Message- each and implement the interface
- Adapt the SpringPayloadExtract (if needed)

Right now, we are unsure about the best solution.

We can also imagine a single class that does handle all GenericTypes in one. And similar to resilience4j.ignoreException, a user can set the types which should be extracted via configuration:
```properties
springwolf:
    ...:
        payloadExtractor:
            types:
                - org.springframework.web.client.HttpServerErrorException
                - java.io.IOException
```

SheheryarAamir:
thank you @timonback. I have one question, how to extract payload type in we are using custom interface instead of `spring message` for example `MyMessage<String>` ?

timonback:
Hi, sorry I forgot to update and good question.

I am not sure if the current state is the final idea before the release. I imagine that the map of classes that can be extracted can be passed via properties configuration by the user.
The map is likely also missing some entries. Function, Consumer and Supplier comes to mind. Are you missing some entries?

SheheryarAamir:
No, I dont think of any more entries other than already mentioned. But giving an option to user to put values in map solve most of the problem.
I also like the `fasterxml` implementation of deserialisers . You can have a look in class `SimpleDeserializers` of `jackson-databind:2.15.2` library

PR #442, opened 2023-11-09
chore: Upgrade to Gradle 8.4

ctasada:
Upgraded to the latest Gradle stable version.

See https://docs.gradle.org/current/userguide/upgrading_version_7.html


timonback (approved):

sam0r040:
Thanks for the Gradle update and the other small fixes!

Issue #427, opened 2023-10-30
Support Avro Schemas and Schema Registry


raphaeldelio:
**Describe the feature request**
When using Springwolf with Avro Models:
- the generated example is generated with a lot of field that should've been ignored.
- the yaml file is not parsable by the AsyncAPI studio
- the UI doesn't display information about the schema registry
- The documentation is not clear on how to publish messages using a Schema Registry

The example message should be:
```
{
  "someString" : "text",
  "someLong" : 0
}
```

Github Repository with project:
https://github.com/raphaeldelio/demo-springwolf

AsyncAPI Studio:
<img width="569" alt="Screenshot 2023-10-30 at 10 26 24" src="https://github.com/springwolf/springwolf-core/assets/25641721/a3511299-ca3e-42b7-9c76-b84737f76f33">
<img width="570" alt="Screenshot 2023-10-30 at 10 26 28" src="https://github.com/springwolf/springwolf-core/assets/25641721/ab71b377-7e8a-470b-99d7-c6f636f95b5b">

Springwolf UI:
<img width="896" alt="Screenshot 2023-10-30 at 10 27 28" src="https://github.com/springwolf/springwolf-core/assets/25641721/8101a8b4-7ff8-4e51-9d10-a54701c015a4">

timonback:
Hi @raphaeldelio 
Thank you for the report, the suggestion and the linked demo repo.

Regarding the questions:
1. We do not have native support for avro yet.
2. True, the refs on payload level are not resolved in the asyncapi UI. This will be resolved when the 3.0 spec is released, which has been postponed.
3. + 4. Schema registries are also not supported yet. Shall we move that to a different enhancement?

Feel free to improve the support as there are likely more people who use avro.
We are happy to assist and connect on Discord.

raphaeldelio:
@timonback Thank you! I will test it when the snapshot is available.

timonback:
Hi @raphaeldelio,
the `SNAPSHOT` version is available now

raphaeldelio:
@timonback just tested it and it works! I tested it with an Avro Model that has nested records as well. Thank you for this. When is this planned to be released?

timonback:
Awesome.
We plan to release a new version at the end of each month, so the next one is at the end of February

timonback:
Thank you very much for the shoutout on LinkedIn!

Just to clarify, Avro is supported as of the current release.
AsyncAPI 3 including the fix for the schema will will be part of the February release.
[...]

PR #503, opened 2023-12-18
feat: Remove avro fields from schema

timonback:


timonback:
Relates to https://github.com/springwolf/springwolf-core/issues/427

timonback:
We discussed and arrived at the following:
- We like to add a SchemaPostProcessor interface
  - It is activated by default through a bean factory (no activation toggle required)
- It can stay in core as long as it does not require new dependencies
  - (We do not add it to kafka as the cloud-stream plugin will use the same logic)
- Indicate through the package name (and README) that the location of the class is experimental and may be moved later - for example to an own add-on.
- topic name: "avro-topic-without-publishing-supported" + Description

PR #516, opened 2023-12-29
Adding kafka protubuf example

alievrenkut:
I added kafka probuf example for springwolf. You can try it by reading README.md

Important bits:

- ObjectMapperConfiguration adds ProtobufModule so Jackson can create correct shemas for protobuf objects also convert them correctly.
- ProtobufPropertiesModule class is copy paste from internet they do not have maven artifact so this is best I can do

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -13,6 +13,7 @@ include(
         'springwolf-examples:springwolf-cloud-stream-example',
         'springwolf-examples:springwolf-jms-example',
         'springwolf-examples:springwolf-kafka-example',
+        'springwolf-examples:springwolf-kafka-protobuf-example',
         'springwolf-examples:springwolf-sns-example',
         'springwolf-examples:springwolf-sqs-example',
         'springwolf-ui',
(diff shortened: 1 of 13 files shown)


timonback:
Thank you @alievrenkut. This example has been a great help.

I improved on it in https://github.com/springwolf/springwolf-core/pull/526 and integrated it into the existing kafka example.
In the meantime, the kafka example will allow to publish protobuf as well as. However users will have to add a custom class to their project: https://github.com/springwolf/springwolf-core/blob/9f69f82c615c325d9986f9fc52ce55c8d8a7e159/springwolf-examples/springwolf-kafka-example/src/main/java/io/github/stavshamir/springwolf/example/kafka/configuration/CustomSpringwolfKafkaProducer.java#L42

---
By the way, the PR to add Optional as an extractable class by default is also created now: https://github.com/springwolf/springwolf-core/pull/525

timonback:
Thank you @alievrenkut for the example.

The PR #526 has been merged to demonstrate how to use protobuf (incl. producer). You will need to add the CustomSpringwolfProducer.

PR #517, opened 2023-12-31
Enhancement: Add More Spec Details to the UI

### LINKED ISSUE #378: Enhancement: Add More Spec Details to the UI
**Description:**
I've noticed that certain Swagger documentation annotations, notably 'minimum' and 'maximum', are present in the specification but are not reflected in the UI. Additionally, payloads with primitive types like 'string' don't have their type explicitly displayed in the UI.

**Steps to Reproduce:**

1. Add annotations like 'minimum' or 'maximum' to your Schema Annotation or use a primitive type as payload.
2. Verify they appear in the specification and note that 'type: string' (or similar) is present.
3. Observe the absence of these details in the UI.

**Expected Behavior:**
Swagger annotations, such as 'minimum' and 'maximum', would be a valuable addition to the UI. It would also be beneficial for the UI to display primitive types explicitly.

**Actual Behavior:**
Annotations are missing, and primitive types are not distinctly represented.

**Suggested Fix:**

Enhance the UI to incorporate Swagger annotations more comprehensively.
Display primitive payload types clearly.
Aim for parity with the display capabilities of the [AsyncAPI React component](https://github.com/asyncapi/asyncapi-react).

**Status:**

- [ ] Const
- [x] Enum
- [x] Items
- [x] MaxItems
- [x] MinItems
- [x] Required
- [x] name: String (default: "")
- [x] title: String (default: "")
- [x] multipleOf: double (default: 0.0)
- [x] maximum: String (default: "")
- [x] exclusiveMaximum: boolean (default: false)
- [x] minimum: String (default: "")
- [x] exclusiveMinimum: boolean (default: false)
- [x] maxLength: int (default: Integer.MAX_VALUE)
- [x] minLength: int (default: 0)
- [x] pattern: String (default: "")
- [ ] maxProperties: int (default: 0)
- [ ] minProperties: int (default: 0)
- [ ] requiredProperties: String[] (default: {})
- [ ] requiredMode: Schema.RequiredMode (default: Schema.RequiredMode.AUTO)
- [x] description: String (default: "")
- [x] format: String (default: "")
- [x] ref: String (default: "")
- [x] nullable: boolean (default: false)
- [x] example: String (default: "")
- [ ] externalDocs: ExternalDocumentation (default: @ExternalDocumentation)
- [ ] deprecated: boolean (default: false)
- [x] type: String (default: "")
- [ ] allowableValues: String[] (default: {})
- [ ] defaultValue: String (default: "")
- [ ] discriminatorProperty: String (default: "")
- [ ] hidden: boolean (default: false)
- [ ] enumAsRef: boolean (default: false)
- [x] types: String[] (default: {})
- [x] exclusiveMaximumValue: int (default: 0)
- [x] exclusiveMinimumValue: int (default: 0)

aerfus:
Primitive types have no payload information displayed,
minimum and maximum values are absent in UI.

**Additional information**
A `type` badge is added to payload header

Resolves #378

### DIFF
--- a/springwolf-add-ons/springwolf-common-model-converters/src/main/java/io/github/stavshamir/springwolf/addons/common_model_converters/converters/monetaryamount/MonetaryAmount.java
+++ b/springwolf-add-ons/springwolf-common-model-converters/src/main/java/io/github/stavshamir/springwolf/addons/common_model_converters/converters/monetaryamount/MonetaryAmount.java
@@ -10,7 +10,7 @@ import java.math.BigDecimal;
 class MonetaryAmount {
 
     @JsonProperty("amount")
-    @Schema(example = "99.99")
+    @Schema(example = "99.99", minimum = "0.1")
     private BigDecimal amount;
 
     @JsonProperty("currency")
--- a/springwolf-core/src/test/java/io/github/stavshamir/springwolf/asyncapi/DefaultAsyncApiSerializerServiceIntegrationTest.java
+++ b/springwolf-core/src/test/java/io/github/stavshamir/springwolf/asyncapi/DefaultAsyncApiSerializerServiceIntegrationTest.java
@@ -19,7 +19,6 @@ import io.swagger.v3.core.converter.ModelConverters;
 import io.swagger.v3.oas.models.media.Schema;
 import io.swagger.v3.oas.models.media.StringSchema;
 import lombok.Data;
-import org.apache.commons.lang3.StringUtils;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -111,7 +110,7 @@ class DefaultAsyncApiSerializerServiceIntegrationTest {
         String actual = serializer.toJsonString(asyncapi);
         InputStream s = this.getClass().getResourceAsStream("/asyncapi/asyncapi.json");
         String expected = new String(s.readAllBytes(), StandardCharsets.UTF_8);
-        assertEquals(StringUtils.deleteWhitespace(expected), StringUtils.deleteWhitespace(actual));
+        assertEquals(expected, actual);
     }
 
     @Test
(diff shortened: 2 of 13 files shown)


timonback (commented):
Hi @aerfus 
I like the code changes (incl ng template) and how the data is presented to the user.

A couple things/questions:
1. To fix the formatting, you can run gradle spotlessaApply
2. the AsyncAPI.json file is generated by Springwolf through the code. There should be a Size annotation for a class property or similar that will result in the minimum/maximum field in the spec.
3. Same thing with the primitives, which you probably added for demo purposes. And then underlying issue being that the ui side has no test coverage. If you want to investigate how to add angular ui tests, you are very welcome. Otherwise, adding a single string Kafka consumer for the Kafka example would be sufficient from my side
4. Since us maintainers also stumble upon the ignoreWhitespaces: Did you had trouble getting the tests running/passing? 

Great work!

aerfus:
Hi @timonback ,

A couple things/questions:

1. Formatting is fixed
2. Done
3. Done. Yes, those examples were for the demonstration. So I've removed them.
4. Yes, my tests are failing without those changes.
I'm using Windows OS and have .gitattributes present. Still the generated fixture has Windows line endings, which results in test failures.

aerfus:
Let me investigate angular ui tests.
For that I'll create a separate issue and PR.

timonback:
@aerfus Can you check if there is a way to assert the schemas, while keeping a line ending?
For example: `assertThat(actualDefinitions).isEqualToNormalizingNewlines(expected);`

Does the ApiSystemTest in the example fail as well? If not, maybe the files json files in core have been checked in with the wrong line ending.

aerfus:
Hi @timonback ,

I've made a little research.
`ApiSystemTest` does no fail as it uses restTemplate, which returns JSON with '\n' line ending.
All the resources involved also have '\n' line ending.
The problem occurs when we use `DefaultAsyncApiSerializerService`.
I've managed to force it using '\n' and tests are now passing.

Should I rewrite tests using `isEqualToNormalizingNewlines`
or changing `CustomPrettyPrinter` is okay?

jenarp:
Hi @aerfus,
this looks good so far, thank you.
What I am missing is to distinguish between exclusive and inclusive max and mins, i.e. bigger equal (>=) vs. bigger (>) and the same with lower.
Do you see a chance to add such behaviour. In Swagger it is possible to distinguish between these annotations: exclusiveMinimum/exclusiveMaximum and just minimum/maximum (which is inclusive by default).

timonback:
@aerfus Let's do the following:

I revert the commit that improves development under windows and we move that to a separate PR. I prefer the idea of `isEqualToNormalizingNewlines` as it only touches test code and not production. To test it, you can also open a PR against your own fork, then I do not have to approve Github actions to run.

Running `git ls-files --eol` shows on my system that (almost) all files are checked in as `lf` as it should be on my system. I am happy to check and debug further with you together.

Why merge now?
It simplifies our preparation for AsyncApi 3.0 and I want to avoid merge/rebase conflicts.

PR #524, opened 2024-01-05
chore: remove deprecated asyncapidocket bean

timonback:
The AsyncApiDocket still exists as a data holder, but is not used as a bean anymore.

I wonder, whether the is the need for this class at all. Alternatively, services can access the SpringwolfProperties directly. But that can be potentially unstable in the long run as well.


springwolf-core/src/main/java/io/github/stavshamir/springwolf/configuration/DefaultAsyncApiDocketService.java
```diff
@@ -73,7 +56,7 @@ private AsyncApiDocket parseApplicationConfigProperties() {
             builder.defaultContentType(configProperties.getDocket().getDefaultContentType());
         }
 
-        return builder.build();
+        docket = builder.build();
```
ctasada, line 59: NBC: I think it's more transparent to `return builder.build()` and then

```
    public AsyncApiDocket getAsyncApiDocket() {
        if (docket == null) {
            docket = createDocket();
        }
        return docket;
    }
``` 
In this way the check and initialization are in the same place, avoiding "hidden" side-effects

ctasada (approved):

PR #541, opened 2024-01-14
Feat/message as component

timonback:
Open question: Which field do we use for referencing message? The id or name field?
My guess is that id should be used for check of duplication in the ChannelMerger and elsewhere. And name is just a display field for the user.

TODO: review if all methods of Reference classes are used


springwolf-core/src/test/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/ChannelMergerTest.java
```diff
@@ -129,28 +130,28 @@ void shouldMergeDifferentMessagesForSameChannel() {
         // given
         String channelName = "channel";
         MessageObject message1 = MessageObject.builder()
-                .messageId("message1")
+                .messageId("messageString")
```
timonback, line 133: marker: id vs name

springwolf-core/src/test/resources/asyncapi/asyncapi.yaml
```diff
@@ -23,25 +23,22 @@ channels:
...
-      - $ref: '#/servers/production'
+      - "$ref": "#/servers/production"
+components:
+  schemas:
+    ExamplePayload:
+      type: object
+      properties:
+        s:
+          type: string
 operations:
```
timonback, line 37: messages is missing here

springwolf-asyncapi/src/main/java/io/github/stavshamir/springwolf/asyncapi/v3/model/channel/ChannelObject.java
```diff
@@ -51,7 +52,8 @@ public class ChannelObject extends ExtendableObject implements Channel {
      * this channel MUST be valid against one, and only one, of the message objects defined in this map.
      */
     @JsonProperty(value = "messages")
-    private Map<String, Message> messages;
+    @Builder.Default
+    private Map<String, MessageReference> messages = new HashMap<>();
```
ctasada, line 56: NBC: `messages` are not mandatory. I don't see any reason for a channel to **not** have an associated message, but based on the spec, it's not a required field.

springwolf-core/src/test/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/ChannelMergerTest.java
ctasada: `messageId` is intended to be used as an `id` on the messages Map references, while the `name` is a 'machine-friendly name'. In most places I used the new `messageId` while in others I used the `name` to keep backward compatibility.

I think it's totally fair to provide a guideline and align the implementation to follow a single approach.

If we consider `name` as 'machine-friendly' it can totally work as a `messageId` and merge both concepts. I do not have a hard opinion there.

ctasada (approved):

springwolf-core/src/test/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/ChannelMergerTest.java
timonback: Thanks. I had a closer look. messageId is used only for the spec.
I sticked to your suggestion to stay backwards compatible and use `name`

PR #543, opened 2024-01-15
Different fixes in the asyncapi v3 spec examples

ctasada:
Some of the used examples to test the code were not passing the AsyncAPI v3 validation process. Here we fix them and leave the code ready for a next iteration of improvements

### DIFF
--- a/springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/annotation/AsyncAnnotationChannelsScanner.java
+++ b/springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/annotation/AsyncAnnotationChannelsScanner.java
@@ -142,7 +142,6 @@ public class AsyncAnnotationChannelsScanner<A extends Annotation>
 
         List<String> servers = AsyncAnnotationScannerUtil.getServers(operationAnnotation, resolver);
         if (servers != null && !servers.isEmpty()) {
-            // FIXME: It was originally operationId, which doesn't exist anymore
             validateServers(servers, operation.getTitle());
             channelBuilder.servers(servers.stream()
                     .map(it -> ServerReference.builder().ref(it).build())
@@ -157,7 +156,6 @@ public class AsyncAnnotationChannelsScanner<A extends Annotation>
     }
 
     private Map.Entry<String, Operation> buildOperation(MethodAndAnnotation<A> methodAndAnnotation) {
-        Operation.OperationBuilder operationBuilder = Operation.builder();
         AsyncOperation operationAnnotation =
                 this.asyncAnnotationProvider.getAsyncOperation(methodAndAnnotation.annotation());
         String operationName = resolver.resolveStringValue(operationAnnotation.channelName());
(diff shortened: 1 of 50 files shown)


springwolf-plugins/springwolf-sns-plugin/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/operationdata/annotation/SnsAsyncOperationBinding.java
```diff
@@ -20,4 +20,10 @@
 public @interface SnsAsyncOperationBinding {
```
ctasada, line 20: Both `protocol` and `endpoint` are mandatory fields for which I don't think we should provide a default value. This decision automatically generates a breaking change. Without a value, the compilation of any code using this annotation will fail.

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/annotation/AsyncAnnotationChannelsScanner.java
```diff
@@ -94,13 +94,7 @@ private Stream<MethodAndAnnotation<A>> getAnnotatedMethods(Class<?> type) {
...
 
     private boolean isInvalidChannel(Map.Entry<String, ChannelObject> entry) {
-        //        Operation publish = entry.getValue().getPublish();
-        //        boolean publishBindingExists = publish != null && publish.getBindings() != null;
-        //
-        //        Operation subscribe = entry.getValue().getSubscribe();
-        //        boolean subscribeBindingExists = subscribe != null && subscribe.getBindings() != null;
-
-        boolean allNonNull = entry.getKey() != null; // && (publishBindingExists || subscribeBindingExists); FIXME
+        boolean allNonNull = entry.getKey() != null;
```
timonback, line 97: I believe a map key can never be null be design.
-> remove this method?

springwolf-plugins/springwolf-cloud-stream-plugin/src/test/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/cloudstream/CloudStreamFunctionChannelsScannerIntegrationTest.java
```diff
@@ -98,30 +101,34 @@ void testConsumerBinding() {
...
 
         ChannelObject expectedChannel = ChannelObject.builder()
                 .bindings(channelBinding)
-                // .publish(operation) FIXME
+                .messages(Map.of(message.getName(), message))
                 .build();
 
-        assertThat(channels).containsExactly(Map.entry(topicName, expectedChannel));
+        Operation expectedOperation = Operation.builder()
+                .action(OperationAction.SEND)
```
timonback, line 122: Nitpicking: Receive action

timonback (approved):
Again, great work!

So good to see an actual sqs/sns binding

springwolf-core/src/main/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/annotation/AsyncAnnotationChannelsScanner.java
ctasada: Yes, I think that would be the best. If in the future we find out that some conditions are generating broken instances of Channel or Operation, we can add back the filter

springwolf-plugins/springwolf-cloud-stream-plugin/src/test/java/io/github/stavshamir/springwolf/asyncapi/scanners/channels/cloudstream/CloudStreamFunctionChannelsScannerIntegrationTest.java
ctasada: I will try to fix it tonight

timonback (approved):

Issue #542, opened 2024-01-15
Markdown support


nschwalbe:
**Describe the feature request**
Markdown support in description fields

**Motivation**
A more structured and formatted documentation. Currently every text is just a line without breaks or paragraphs. This is very difficult to read for some longer descriptions.

**Technical details**
Maybe include angular ngx-markdown library

**Describe alternatives you've considered**
-

timonback:
Hi @nschwalbe,
Thank you for the suggestion.

Using Commonmark (markdown) for the structured text is great, as it is suggested the AsyncApi spec. See https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md#fixed-fields-1

The description field of the info object is a great start.
Do you use a long description for the Channels & Schemas as well?

Feel free to get started on it (or anyone how wants to contribute it)

nschwalbe:
Yes, exactly. We want to describe the info, channel and schema in more detail.
My colleague is preparing a PR for the ngx-markdown library which uses marked which is nearly complete commonmark 3.0 compliant: https://github.com/markedjs/marked/discussions/1202#discussioncomment-4192078

timonback:
Thank you for the contribution, it is part of the latest release.

Feel free to reopen this issue if there is still something missing.

PR #549, opened 2024-01-18
Feat/adds markdown support

robert-henke:
Hi, this pull request considers the requested markdown support described in https://github.com/springwolf/springwolf-core/issues/542 . It uses ngx-markdown with enabled mermaid and prismjs support. It allows to use markdown syntax in all description fields.


timonback:
Thank you @robert-henke for this contribution!

To showcase this on our demo website, it is best to also update at least the description of the examples/kafka-example application as well. The new description could be something like the existing one, new line, ˋmarkdownˋ syntax is _supported_ as well. (I am open for creative suggestions as well)
Possibly showcase also a line of code as you included prism?
Can you do that?

Do you know how prism compares to highlightjs, which we already use to render json?

Mermaid is a larger dependency, so I am hesitant to include it just because ngx-markdown supports it. Do you have a specific case in mind, where you want to visualize a diagram in the info section and/or the channels?

robert-henke:
Hi @timonback  I added some example texts for the description fields. Let me know if the provided examples meet your expectations.

Considering the mermaid lib: We would like to add some flow charts and / or sequence diagrams to document message flows between services. However, this is not essential for us, as we could simply include an image with the diagram. It would just be nice to have a single source of code for further adaptations of diagrams etc.

The prismjs lib is used by ngx-markdown to highlight code fragments inside code blocks. I could not find a (documented) way to use highlightjs for this purpose. That is why I included the prismjs dependency as well. The lib is used by markdown by simply adding the required language as part of a text block syntax in markdown (e.g. `\`\`\`json `). I included an example for code highlighting as well.

robert-henke:
Hi @timonback I adapted the json file as proposed, I also had to add some minor adjustments to the `OpenApiGeneratorSystemTest`. The kafka example test target should run without errors now. Let me know of any other required adjustments.

From my (limited) understanding of prismjs and the current use of highlightjs I could imagine that it is absolutely possible to realize those use-cases with prismjs as well.

timonback:
We decided to remove mermaid support as the dependency because we are unsure about the use-case. To us, it seems like that you try to document something that should be located in other tools (Confluent, Backstage, ApiHub) that gives a broad overview over all services - not a single one).

Also, the dependency is rather large and as soon as we add it, we will have to support it forever.

---

Still, thank you for this very nice addition and the work you have put into it. It will be part of the next release 🚀

PR #567, opened 2024-01-26
Chore/async api 3 clean up

timonback:


springwolf-asyncapi/src/main/java/io/github/stavshamir/springwolf/asyncapi/v3/model/AsyncAPI.java
```diff
@@ -82,6 +76,13 @@ public class AsyncAPI extends ExtendableObject {
     @JsonProperty(value = "defaultContentType")
     private String defaultContentType;
 
+    // TODO: moved only for testing purposes - revert later
```
timonback, line 79: TODO: move

timonback:
@ctasada This introduces a `SwaggerSchemaUtil` to map from swagger to AsyncApi schema. I left a note there for the `discriminator` field, so that it can be mapped correctly.

Issue #160, opened 2023-03-31
discriminator field generated does not match asyncapi spec


TabraizChel:
**Describe the bug**
If an object has a polymorphic field when the json schema is generated for the object the discriminator property follows the openapi v3 convention (discriminator field is an object) whereas openapi spec wants the discriminator property as a string. This results in an invalid async api doc. 

**Dependencies and versions used**
`springwolf-kafka` version `0.10.3`.
`spring-boot` version `3.0.5`.

**Code example**
```
package com.example.demo.consumer;


import com.fasterxml.jackson.annotation.JsonSubTypes;
import com.fasterxml.jackson.annotation.JsonTypeInfo;

@JsonTypeInfo(
        use = JsonTypeInfo.Id.NAME,
        property = "type"
)
@JsonSubTypes(value = {
        @JsonSubTypes.Type(value = Dog.class, name = "dog"),
        @JsonSubTypes.Type(value = Cat.class, name = "cat"),
})
public interface Pet {

    public String getType();
}

\\\\\\\\\\\\\\\\\\\\\\\\\\\

package com.example.demo.consumer;

public record Cat() implements Pet {
    @Override
    public String getType() {
        return "cat";
    }
}

\\\\\\\\\\\\\\\\\\\\\\\\\\\

package com.example.demo.consumer;

public record Dog() implements Pet {
    @Override
    public String getType() {
        return "dog";
    }
}

\\\\\\\\\\\\\\\\\\\\\\\\\\\

package com.example.demo.consumer;

public record MyEvent(
  Pet pet
)  {
}

\\\\\\\\\\\\\\\\\\\\\\\\\\\

package com.example.demo.consumer;

import org.springframework.kafka.annotation.KafkaListener;
import org.springframework.stereotype.Service;

@Service
public class Listener {
    @KafkaListener(topics = "saas-bifrost-response")
    public void process(MyEvent event) {

    }
}
```
This results in a Pet portion of the schema being:
```
      "Pet" : {
        "type" : "object",
        "properties" : {
          "type" : {
            "type" : "string",
            "exampleSetFlag" : false,
            "types" : [ "string" ]
          }
        },
        "example" : {
          "type" : "string"
        },
        "discriminator" : {
          "propertyName" : "type"
        },
        "exampleSetFlag" : true
      }
```
note the  `discriminator` field. For asyncapi spec this should just be:

```
"discriminator": "type"
```
[...]

PR #569, opened 2024-01-28
fix: handle AsyncApi v3 maximumValue field

timonback:
Swagger generates OpenApi schemas which are not 100% compatible to AsyncApi

### DIFF
--- a/springwolf-core/src/test/resources/asyncapi/asyncapi.json
+++ b/springwolf-core/src/test/resources/asyncapi/asyncapi.json
@@ -43,7 +43,6 @@
   "components": {
     "schemas": {
       "ExamplePayload": {
-        "type": "object",
         "properties": {
           "s": {
             "type": "string"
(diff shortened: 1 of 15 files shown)


springwolf-core/src/test/resources/asyncapi/asyncapi.json
```diff
@@ -43,7 +43,6 @@
   "components": {
     "schemas": {
       "ExamplePayload": {
-        "type": "object",
```
timonback, line 46: The PostProcessor fixes this, which is not part of this test

timonback:
@aerfus We stumbled upon the maximumValue field in the AsyncApi 3 spec that you also mentioned in the other PR. A review - if you find time - is highly appreciated.

aerfus:
Hi @timonback ,
I've reviewed and have no concerns.
Do you mean my notice on exclusiveMinimum(Maximum) became a number,
not boolean?
Is so, should we rewrite range component logic?

ctasada (approved):
I think we can apply the same approach for https://github.com/springwolf/springwolf-core/issues/160 WDYT?

Issue #563, opened 2024-01-25
Add support for Payload Kafka Key Types


Sonic-Rage:
We utilize an Object as a Key to send events with Kafka.   It would be great if the scanner could recognize objects used for Kafka Keys to add the schemas automatically to message binding. 

Here is a sample of a listener that we use with Spring Kafka

 @KafkaHandler
 public void consumeMessage(
            @Headers MessageHeaders headers,
            @Header(KafkaHeaders.RECEIVED_KEY) KeyObject key,
            @Payload(required = false) PayloadObject payload) {

ctasada:
Hi @Sonic-Rage I was looking into this ticket. Would be fair to assume that, given something like

```java
    @KafkaHandler
    public void receiveExamplePayload(
            @Headers MessageHeaders headers,
            @Header(KafkaHeaders.RECEIVED_KEY) String key,
            @Payload(required = false) ExamplePayloadDto payload) {
        log.info("Received new message in {}: {}", TOPIC, payload.toString());
    }
```

Should generate 
```yaml
        "message": {
          "schemaFormat": "application/vnd.oai.openapi+json;version=3.0.0",
          "name": "io.github.stavshamir.springwolf.example.kafka.dtos.AnotherPayloadDto",
          "title": "AnotherPayloadDto",
          "payload": {
            "$ref": "#/components/schemas/io.github.stavshamir.springwolf.example.kafka.dtos.AnotherPayloadDto"
          },
          "headers": {
            "type": "object",
            "properties": {
                "id": {
                   "type": "string"
                 },
                "kafka_receivedMessageKey": {
                   "type": "string"
                 },
                "timestamp": {
                   "type": "string"
                 },
          },
          "bindings": {
            "kafka": {
              "bindingVersion": "0.4.0"
            }
          }
        }
```
We are heavily working on the next release, but we can check how to improve this case. Can you confirm that would be your expectation?

Sonic-Rage:
@ctasada Sorry it took me a little bit to go through the Async spec some.   The key for Kafka doesn't appear to be well documented from what I can tell so hopefully I'm not making too many assumptions. 
My use case involves all three( Kafka headers, key and payload)  Headers for traceability, and objects in Key for filtering and partioning Kafka events as well as payload objects.  

The Readme here shows the key as https://github.com/asyncapi/bindings/blob/master/kafka/README.md#message-binding-object gets configured as a message binding.  

[...]

PR #579, opened 2024-01-31
feat: Added support for Spring Messaging @Header annotations

ctasada:
Fixes issue #563

Now, when a method is annotated with the `@Header` or `@Headers` annotation, the payload marked with `@Payload` will also include the defined headers.

Known limitations:
* If a @Payload is indicated in different places, the AsyncAPI specification will include any correlated @Header
* When using `@Headers` we cannot provide further information


sam0r040:
Superseded by https://github.com/springwolf/springwolf-core/pull/787

PR #593, opened 2024-02-09
feat(core): Skip schema post processor when previous post-processor r…

timonback:
…emoved the current schema

Example: an avro schema is removed. Then the next post-processor (example generator) does not need to be called. Also, it avoids the error message due to unresolvable avro schemas


timonback:
Relates to the second issue of https://github.com/springwolf/springwolf-core/pull/582

PR #615, opened 2024-02-24
fix: override schema payload type to string for non json contentType

timonback:
Ensures that the AsyncApi studio does render the asyncapi.json file

This PR is open for discussion, especially the handling of `ListWrapper`.

Goal: Release a springwolf version, whose asyncapi.json files can be rendered by AsyncApi Studio.


timonback:
If we merge this, the adjusted test in asyncapi mapper in ui can be reverted

PR #634, opened 2024-03-02
(feat): Decoupled Binding annotations

ctasada:
Decouple the binding annotations.

The annotations to document the bindings were only available via the different springwolf-plugins, and were forcing to import multiple extra dependencies.

Those annotations are now published in it's own module without extra dependencies, so they can be use for documentation goals. If you want to produce/consume events, you still need to use the adequate `springwolf-plugin`

* `springwolf-bindings:springwolf-sqs-binding`
    * `@SqsAsyncOperationBinding`
    * `@SqsAsyncQueueBinding`
* `springwolf-bindings:springwolf-sns-binding`
    *  `@SnsAsyncOperationBinding`
    *  `@SnsAsyncOperationBindingIdentifier`
* Pending
    * `@AmqpAsyncOperationBinding`
    * `@JmsAsyncOperationBinding`
    * `@KafkaAsyncOperationBinding`


timonback:
Hi @ctasada,

great idea, however the timing is unfortunate as we just released 1.0.0, which allowed us to easily discuss and including breaking changes.

---

We discussed more in detail and focused on the argument of the dependencies that you mention as well.

We believe that the current Springwolf structure using the plugin mechanism is very powerful. 
A plugin currently adds 3 aspects:
1. The detection of the spring listener annotation (only additional dependency, i.e. `spring-kafka`)
2. The publishing endpoint (i.e. the `spring-web` dependency that is present in core as well)
3. The binding (focus of this PR)

Moving the bindings into the core, will result in many spring beans that are being created and not used in core.
More importantly, it might result in a bloated core as more and more protocol specific functionality is added/moved. The plugins are a nice way, to keep those out of core.

---

We thought about an alternative: New maven artifacts in the form of i.e. `<plugin>-binding`

The plugins do have a direct dependency on these artifacts, so existing users do not need to add another artifact. The new artifact is for the "power users", who want a pure `core` plus `binding` functionality.

To avoid breaking changes, the Spring `aliasFor` annotation marker can be used in case the annotation is moved to a different package - again to avoid breaking changes.

Side note: We can add/move the `generic-bindings` add-on as well.

---

Is your main motivation the dependencies only, or is there something else as well?

ctasada:
Hey @timonback 

My main motivation for this change is the injection of extra dependencies. For example, I have a service using SNS, but it's directly using the AWS client and I don't want to inject extra dependencies just for documentation.

I understand the concerns and, in fact, the creation of a `<plugin>-binding` would work.

With this approach in mind, I can refactor this MR:

* Add the new `plugin-binding` artifacts
* Move the annotations to the new bindings, keeping the packaging

Would that work? Alternatively we can duplicate the annotations and mark the current ones as deprecated.

timonback:
Sounds great, thanks for the clarification @ctasada.

We do have an own package namespace for each maven artifact, so the new annotation would contain the package name `<plugin>-binding` and the existing annotation would reference the new one.

Feel free to go ahead - either with one plugin at first (and we review) or all together.

ctasada:
@timonback I refactored the whole MR and migrated **only** the SQS bindings.

Is this the approach you had in mind?

.github/workflows/springwolf-plugins-bindings.yml
```diff
@@ -0,0 +1,56 @@
+name: springwolf-plugins-bindings
```
timonback, line 1: Assuming that we only add bindings for existing plugins, I can imagine that we add it to the existing plugin pipeline

springwolf-plugins/springwolf-sqs-plugin-binding/src/main/java/io/github/springwolf/plugins/sqs/configuration/SpringwolfSqsBindingAutoConfiguration.java
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: Apache-2.0
+package io.github.springwolf.plugins.sqs.configuration;
```
timonback, line 2: For the `1.0.0` release, each gradle module has its own package namespace.

Idea:
```suggestion
package io.github.springwolf.bindings.sqs.configuration;
```

Similarly, I wonder whether the gradle module should be outside of the `springwolf-plugins` folder and have its own `springwolf-bindings` folder.

I am happy to discuss and also get feedback from @sam0r040

timonback (commented):
Looks good.

My main concern is the bindings location within the project (which folder + namespace).

---

I just read about the `AliasFor` marker on annotations. I am not sure if that can be used for the full annotation - to avoid the breaking change for users. But that can be an item for the next step.

springwolf-plugins/springwolf-sqs-plugin-binding/src/main/java/io/github/springwolf/plugins/sqs/configuration/SpringwolfSqsBindingAutoConfiguration.java
ctasada: @timonback I don't think we can use `@AliasFor` since it doesn't support `@interface`

I will go with a full duplication and add a @deprecation to the existing annotations (timing is kind of bad)

I create its own pipeline for the `bindings` because there is no guarantee parity, but then it may make more sense to remove the `plugin` reference to the artifact.

ctasada:
@timonback @sam0r040 I did a new refactor.

Now the code is duplicated and the `binding` artifacts are in its own pipeline, so we have 100% decoupling from plugins.

I decided not to go with deprecation because it's really noisy, but I will fix it later

springwolf-plugins/springwolf-sqs-plugin-binding/src/main/java/io/github/springwolf/plugins/sqs/configuration/SpringwolfSqsBindingAutoConfiguration.java
sam0r040: Not sure where we continue our discussion, but I'll comment here. I think structure wise it looks good now, but the duplication worries me a bit. Did you think about changing the annotations in the plugins to meta annotations which would just "redirect" to the new ones? If our resolution mechanism can handle them, thats what I would vote for.

ctasada: @sam0r040 I have tried using the meta annotations approach but I'm facing some limitations that I don't see how to address (check the error with the sns-example).

timonback: Yes @ctasada , neither do I see a way to resolve the annotation in annotation issue.
Didn't expect that to be honest...

I guess the `AliasFor` approach won't be possible then and we have to go with breaking (import) changes -> removing the annotations for the `plugin` and keeping them in the `binding` only.

In terms of structure, the PR looks good to me. I am positive that it can be part of the next release.

PR #649, opened 2024-03-13
fix(ui): Make edit header view available by default in ui

victorlev01:
Hello! Thanks for great project!
submitted issue with explaination https://github.com/springwolf/springwolf-core/issues/650

Problem - no schema for headers without schema in Kafka Listeners (HeadersNotDocumented schema).

Scenario - user wants to send custom headers in springwolf kafka ui, headers are not defined in schema and changes in internal implementation.

Now there are two workarounds to this:

1) Define headers for schema using annotations
2) Add example to schema HeadersNotDocumented

image

### DIFF
--- a/springwolf-ui/src/app/components/channels/channel-main/channel-main.component.html
+++ b/springwolf-ui/src/app/components/channels/channel-main/channel-main.component.html
@@ -35,7 +35,7 @@
             "
           ></textarea>
         </div>
-        <div [hidden]="!(headersExample?.lineCount > 1)" class="flex-column">
+        <div class="flex-column">
           <h4>Header</h4>
           <textarea
             spellcheck="false"


timonback:
Hi @victorlev01 , thank you for the contribution.

Just to clarify, the issue is not that publishing does not work.
Instead in the case of HeadersNotDocumented, the header input box is not shown in the ui, leaving the user unable to modify the headers?

victorlev01:
Hi @timonback , thanks for quick response.

Yes, the issue is that the header input box is not shown in the ui. 
So user cannot change headers there.

On controller side everything works as expected with HeadersNotDocumented.
Listener receives and can process them.

timonback:
Perfect, thanks for the clarification.

I'll merge it right away and it will be part of the next release 🎉

Issue #651, opened 2024-03-14
Missing schemaSetting in case of googlepubsub protocol


SheheryarAamir:
**Describe the bug**
When using googlepubsub as protocol using version 1.0.0 generated asynapi(3.0.0) docs does not add `schemaSetting` under bindings->googlepubsub, which is required for passing asynapi(3.0.0) validation. 

**Dependencies and versions used**
spring-wolf v1.0.0 (CloudStream)

**Code example**
After some initial investigation I found out that class `CloudStreamFunctionChannelsScanner` in method `buildMessageBinding` always map `EmptyMessageBinding` which is wrong.

**Stack trace and error logs**
`googlepubsub` property must have required property `schemaSettings`

timonback:
Thank you @SheheryarAamir for the feature request and contribution, it has been published now.

We highly appreciate if you spare some minutes to update the documentation as you have the knowledge of the google pubsub bindings, for example here: https://github.com/springwolf/springwolf.github.io/blob/master/docs/configuration/documenting-bindings.md

PR #652, opened 2024-03-14
(fix): Google pubsub protocol schema setting - cloudsteam

SheheryarAamir:
When using googlepubsub as protocol generated asynapi docs does not add schemaSetting under bindings->googlepubsub, which is required for passing asynapi validation.

Detailed are mentioned in below reported issue
Issue #651


springwolf-plugins/springwolf-cloud-stream-plugin/src/main/java/io/github/springwolf/plugins/cloudstream/asyncapi/scanners/channels/CloudStreamFunctionChannelsScanner.java
```diff
@@ -96,8 +100,18 @@ private Map<String, MessageBinding> buildMessageBinding() {
         return Map.of(protocolName, new EmptyMessageBinding());
     }
 
-    private Map<String, ChannelBinding> buildChannelBinding() {
+    private Map<String, ChannelBinding> buildChannelBinding(Annotation annotation) {
```
timonback, line 103: This is the right place for the change :)

For the other scanners, we have written some utility to extract the i.e. `@KafkaAsyncOperationBinding` using an OperationBindingProcessor.

In this method, a call to the utility (adapted to channelBinding) is sufficient
```java
Map<String, OperationBinding> operationBinding =
                AsyncAnnotationUtil.processOperationBindingFromAnnotation(method, operationBindingProcessors);
Map<String, OperationBinding> opBinding = operationBinding != null ? new HashMap<>(operationBinding) : null;
```
Source: AsyncAnnotationScanner

springwolf-plugins/springwolf-cloud-stream-plugin/src/main/java/io/github/springwolf/plugins/cloudstream/annotation/GooglePubSubSchemaSetting.java
```diff
@@ -0,0 +1,35 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(value = {ElementType.METHOD})
+@Inherited
+public @interface GooglePubSubSchemaSetting {
```
timonback, line 15: I'd like to keep protocol-specific annotations out of core and cloudstream.

@ctasada is working https://github.com/springwolf/springwolf-core/pull/634 to introduce an own gradle module for each binding.
This seems highly suitable for the google pubsub bindings as well.
There the `@AsyncGooglePubSubChannelBinding` and `AsyncGooglePubSubSchema` annotation can be put together.

This has the benefit that also other plugins can use the pubsub binding and also the cloudstream and pubsub is not intertwined.

springwolf-plugins/springwolf-cloud-stream-plugin/src/main/java/io/github/springwolf/plugins/cloudstream/annotation/GooglePubSubSchemaSetting.java
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * Mapped to {@link GooglePubSubSchemaSettings#getFirstRevisionId()}
+     */
+    String firstRevisionId() default "";
+
+    /**
+     * Mapped to {@link GooglePubSubSchemaSettings#getLastRevisionId()}
+     */
+    String lastRevisionId() default "";
```
timonback, line 34: 👍 For using the original binding as defined in https://github.com/asyncapi/bindings/tree/master/googlepubsub

springwolf-plugins/springwolf-cloud-stream-plugin/src/main/java/io/github/springwolf/plugins/cloudstream/asyncapi/scanners/common/FunctionalChannelBeanBuilder.java
```diff
@@ -22,35 +24,37 @@ public class FunctionalChannelBeanBuilder {
 
     public Set<FunctionalChannelBeanData> fromMethodBean(Method methodBean) {
         Class<?> returnType = methodBean.getReturnType();
-
+        Annotation schemaSetting = methodBean.getAnnotation(GooglePubSubSchemaSetting.class);
         if (Consumer.class.isAssignableFrom(returnType)) {
             Class<?> payloadType = getReturnTypeGenerics(methodBean).get(0);
-            return Set.of(ofConsumer(methodBean.getName(), payloadType));
+            return Set.of(ofConsumer(methodBean.getName(), payloadType, schemaSetting));
```
timonback, line 30: We might have to pass in the complete method now...

springwolf-plugins/springwolf-cloud-stream-plugin/src/test/java/io/github/springwolf/plugins/cloudstream/asyncapi/scanners/channels/CloudStreamFunctionChannelsScannerIntegrationTest.java
```diff
@@ -451,5 +451,11 @@ public Function<String, Integer> testFunction() {
         public Function<KStream<Void, String>, KStream<Void, Integer>> kStreamTestFunction() {
             return stream -> stream.mapValues(s -> 1);
         }
+
+        @Bean
+        @GooglePubSubSchemaSetting(encoding = "BINARY", name = "project/test")
+        public Consumer<String> testPubSubConsumer() {
+            return System.out::println;
+        }
```
timonback, line 459: This test configuration is pulled into every spring boot test

I assume that you only want to have a method that can be used in the CloudStreamFunctionChannelsBindingSchemaSettingIntegrationTest.
A more lightweight approach has been to define a bean within the test class itself, as done: SpringAnnotationMethodLevelChannelsScannerIntegrationTest

The scanner might need some adaption to allow scanning for a specific method (untested)
```java
@Override
    public Map<String, ChannelObject> scan() {
        List<Map.Entry<String, ChannelObject>> channelObjects = beanMethodsScanner.getBeanMethods().stream().flatMap(this::scan).toList();
        return ChannelMerger.mergeChannels(channelObjects);
    }

    public Stream<Map.Entry<String, ChannelObject>> scan(Method method) {
        return functionalChannelBeanBuilder.fromMethodBean(method).stream()
                .filter(this::isChannelBean)
                .map(this::toChannelEntry);
    }
```

timonback (changes requested):
Great that you start working on supporting GooglePubSub bindings.

I left some comments regarding structure. In case you need some assistance with the bindings module, lets hop into the Discord chat together with @ctasada to coordinate (we can also move the files at the end of the PR).

(I usually do not review PR as long as they are in the draft state. Thanks for the ping in Discord.)

springwolf-plugins/springwolf-cloud-stream-plugin/src/main/java/io/github/springwolf/plugins/cloudstream/annotation/GooglePubSubSchemaSetting.java
SheheryarAamir: Yes, I believe this will provide us with an opportunity to redesign the `getProtocolName()` method as well. Rather solely relying on extracting protocol information from the first server definition via `docket.getServers()`, we can also consider setting the protocol based on annotations. This approach allows us to specify various protocols for different methods.

ctasada: @SheheryarAamir I just merged https://github.com/springwolf/springwolf-core/pull/655 I think you can follow the a similar approach for this binding.

SheheryarAamir: sure, thank you

timonback (approved):

timonback:
Thank you @SheheryarAamir for the contribution.

It will be part of the next release 🎉

PR #655, opened 2024-03-16
(feat): Decoupled Binding annotations

ctasada:
Decouple the binding annotations.

The annotations to document the bindings were only available via the different springwolf-plugins, and were forcing to import multiple extra dependencies.

Those annotations are now published in it's own module without extra dependencies, so they can be use for documentation goals. If you want to produce/consume events, you still need to use the adequate `springwolf-plugin`

* `springwolf-bindings:springwolf-sqs-binding`
    * `@SqsAsyncOperationBinding`
    * `@SqsAsyncQueueBinding`
* `springwolf-bindings:springwolf-sns-binding`
    *  `@SnsAsyncOperationBinding`
    *  `@SnsAsyncOperationBindingIdentifier`
* Pending
    * `@AmqpAsyncOperationBinding`
    * `@JmsAsyncOperationBinding`
    * `@KafkaAsyncOperationBinding`


ctasada:
@timonback @sam0r040  I created a new MR with **only** the new binding artifacts. This means that code is duplicated. I think it's something we can address in the future.

Basically I see the next options:

* Do nothing (no binding artifacts): I don't like this option, since, at least in my case, is limiting to document some projects
* Create the bindings: Implemented here. Introduces some code duplication
* Create the bindings, but keeping the `plugin` package names: We get the decoupled bindings, and we keep backward compatibility, but we break the uniqueness of the namespaces.
* Create the bindings with meta annotations: Its what I tried in the other MR, but didn't manage to find a way to make it work.

.github/workflows/springwolf-plugins-bindings.yml
timonback, line 1: Duplicate pipeline?

timonback (approved):

timonback:
Hi @ctasada,
I agree with your approach here. You found an issue in the usage of Springwolf, so lets fix it.

At this point, I am fine to go with the code duplication (as option 1 does not help you and option 4 didn't work with the technology we use).

To unblock you and get this in, I am fine with merging it now.
And we continue the discussion on migrating the original annotations over to the new bindings module.

Issue #589, opened 2024-02-06
Add support for Kotlin Serialization


ctasada:
Springwolf doesn't support the generation of AsyncAPI components when used with Kotlin

Example:
```
@Serializable
data class Period(
    @SerialName("start_date")
    val startDate: String,
    @SerialName("period_end")
    val endDate: String,
)
```

The outcome looks like:
```
    Period:
      required:
        - endDate
        - startDate
      type: object
      properties:
        endDate:
          type: string
        startDate:
          type: string
```

Ignoring the `@SerialName` annotation.

timonback:
Hi @ctasada ,

springwolf uses swagger to get the property names.

Even though I use kotlin myself, adding support for one 'external' annotations opens the door to add support for many more annotations.
Besides, iterating through the class hierarchy to identify all annotations and patch the properties names can be tricky.

Multiple alternatives come to mind:
1. Get support for the `@SerialName` annotation in swagger itself.
2. Create some kind of meta-annotation/code-generation/compiler plugin that adds for each `@SerialName` annotation the swagger `@JsonProperty`/`@Schema` annotation.

Do you have a simple solution in mind?

ctasada:
Agree, I was thinking about maybe something like a Swagger ModelConverter that supports the annotation.

I found this old ticket in SpringDoc which we may use as reference: https://github.com/springdoc/springdoc-openapi/issues/1226

timonback:
Thank you @ctasada for the contribution.
It is released now.

As discussed, we mark it as beta for now.
Feel free to update the documentation - as this becomes stable - to share the effort you put into it to better support Kotlin.

https://github.com/springwolf/springwolf.github.io/blob/master/docs

PR #656, opened 2024-03-16
[feat] Added Kotlinx Serialization support

ctasada:
Added basic support for KotlinX Serialization classes.

This new module serializes Kotlin `@Serializable` classes respecting the `@SerialName` annotation and types.

Solves https://github.com/springwolf/springwolf-core/issues/589


timonback (commented):
I like this implementation as it is a lot more compact and an add-on module seems appropriate.

Can you add a (small) test case for useFqn=true (the name field is sufficient)?

ctasada:
@timonback done.

I will keep iterating over next PRs, but I think it's good enough to go, at least as a `beta`

springwolf-add-ons/springwolf-kotlinx-serialization-model-converter/src/main/kotlin/io/github/springwolf/addons/kotlinx_serialization_model_converter/converter/KotlinxSerializationModelConverter.kt
```diff
@@ -0,0 +1,143 @@
...
+            propertySchema = ArraySchema()
+            val value = (property.returnType.javaType as ParameterizedType).actualTypeArguments[0]
+            propertySchema.items = resolveRefSchema(value, context)
+            propertySchema.uniqueItems = true
+        } else if (property.returnType.jvmErasure.java.isAssignableFrom(Map::class.java)) {
+            val mapType = (property.returnType.javaType as ParameterizedType)
+            val value = mapType.actualTypeArguments[1]
+            propertySchema = MapSchema().additionalProperties(context.resolve(AnnotatedType(value)))
+        } else if (property.returnType.jvmErasure.java.isAssignableFrom(Byte::class.java)) {
+            propertySchema = IntegerSchema()
```
sam0r040, line 82: Why do we need to convert Byte to Integerschema? https://www.asyncapi.com/docs/reference/specification/v3.0.0#dataTypeFormat looks like StringSchema with format byte would be the correct schema type and its also what would be generated by default...

sam0r040:
Hi @ctasada,

we had a look at the changes and like it a lot more than the one that you have shown us before. We like to see:

- More simplified code by using more Kotlin features unless we missed something
- More fine grained and isolated test cases that make understanding and debugging easier

We created https://github.com/ctasada/springwolf-core/pull/3 as as suggestion for both and postpone the release till Wednesday (27.3.)  to give you some time to review the suggestion. Tests can be refactored later, but we would like to make the production code simpler.

From our side it would be sufficient to just merge the PR https://github.com/ctasada/springwolf-core/pull/3 and then we can merge this PR without the need for further changes for the march release.

springwolf-add-ons/springwolf-kotlinx-serialization-model-converter/src/main/kotlin/io/github/springwolf/addons/kotlinx_serialization_model_converter/converter/KotlinxSerializationModelConverter.kt
ctasada: @sam0r040 good catch. It's now fixed

ctasada:
@sam0r040 @timonback I merged the suggested improvements, fixed the `byte` bug and refactor the tests a bit.

Hopefully we are good to go

timonback (approved):
Looks great @ctasada and looking forward to the next improvement to move it beyond the beta status.

PR #669, opened 2024-03-24
test: add asyncapi.yaml test in ApiIntegrationTest

pdalfarr:
Adding a test for asyncapi.yaml

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ plugins {
     id 'ca.cutterslade.analyze' version '1.9.1'
     id 'io.spring.dependency-management' version '1.1.5' apply false
     id 'org.springframework.boot' version '3.2.5' apply false
-    id 'org.owasp.dependencycheck' version '9.1.0'
+    id 'org.owasp.dependencycheck' version '9.2.0'
     id 'com.diffplug.spotless' version '6.25.0'
     id 'com.bmuschko.docker-spring-boot-application' version '9.4.0' apply false
 }


springwolf-examples/springwolf-amqp-example/src/test/resources/asyncapi.yaml
```diff
@@ -0,0 +1,264 @@
...
+          examples:
+            - 5
+        someString:
+          type: string
+          description: Some string field
+          examples:
+            - some string value
+      description: Example payload model
+      examples:
+        - localDate: "2015-07-20"
```
pdalfarr, line 150: generated yaml does contain `localDate: 2015-07-20` (without double quotes) but this is reported as an error by studi.asyncapi.com
This does not necessarily means this is actually an error, given the fact that next line ( `localDateTime: 2015-07-20T15:49:04-07:00` )is  NOT reported as an error by studio.asyncapi.com ....

springwolf-examples/springwolf-amqp-example/src/main/java/io/github/springwolf/examples/amqp/dtos/AnotherPayloadDto.java
```diff
@@ -19,5 +19,5 @@ public class AnotherPayloadDto {
     private String foo;
 
     @Schema(description = "Example field", requiredMode = REQUIRED)
-    private ExamplePayloadDto example;
+    private GenericPayloadDto<ExamplePayloadDto> example;
```
pdalfarr, line 22: Testing generic payload: generated yaml seems to be good: I was not able to reproduce the "?" issue.
Maybe it's because my 'old SpringBoot 2.x' app comes with old dependencies, which are responsible for the problem.
And in SpringBoot 3.x applications, new version of these dependencies are used -> no more bug.

springwolf-examples/springwolf-amqp-example/src/test/resources/asyncapi.yaml
```diff
@@ -0,0 +1,264 @@
...
+          localDateTime: 2015-07-20T15:49:04-07:00
```
pdalfarr, line 151: Here, I did not changed the generated code, i.e., I did not add double-quotes around 2015-07-20T15:49:04-07:00 as studio.asyncapi does not complain about missing double quotes here

PR #672, opened 2024-03-25
(chore) Updated AsyncAPI Kafka binding to version 0.5.0

ctasada:
As per https://github.com/asyncapi/bindings/tree/master/kafka

### DIFF
--- a/springwolf-asyncapi/src/main/java/io/github/springwolf/asyncapi/v3/bindings/kafka/KafkaChannelBinding.java
+++ b/springwolf-asyncapi/src/main/java/io/github/springwolf/asyncapi/v3/bindings/kafka/KafkaChannelBinding.java
@@ -53,5 +53,5 @@ public class KafkaChannelBinding extends ChannelBinding {
      */
     @Builder.Default
     @JsonProperty("bindingVersion")
-    private String bindingVersion = "0.4.0";
+    private String bindingVersion = "0.5.0";
 }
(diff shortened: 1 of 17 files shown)


ctasada:
@timonback maybe you know what's happening here?

Looking at `npm ls --all` I see

```
├─┬ @asyncapi/parser@3.0.10
│ ├─┬ @asyncapi/specs@6.5.3
│ │ └── @types/json-schema@7.0.12
```

Looking at https://github.com/asyncapi/spec-json-schemas/releases/tag/v6.5.3 The spec 6.5.3 added support for Kafka 0.5.0 binding version, but the tests are failing saying that the version is not valid.

timonback:
Hi @ctasada 
I do not fully understand how the parser works, but as itself does not contain spec files, I checked other places.

Kafka 0.5.0 was recently added in https://github.com/asyncapi/spec-json-schemas/pull/501/files and possible forgotten to be added [here](https://github.com/asyncapi/spec-json-schemas/blame/bee34d8569a18ee2117b4f2b0529a4e883e73638/definitions/3.0.0/serverBindingsObject.json#L53) as well?

I haven’t deeply checked and it is just the first thing I stumble upon. There can be other places as well, or this isn’t the right place at all.

In case you have an idea to get better error message that show at which path the validation fails, let us know/improve the code.

ctasada:
https://github.com/asyncapi/parser-js/releases/tag/v3.0.11 is released. Once Dependabot updates this dependency I will rebase this branch

timonback (approved):

PR #679, opened 2024-03-31
feat: Added support for arrays to the Generic Binding

ctasada:
When using the Generic Binding, it was not possible to define an array value. Here we add support for it.

Example: "key=[value1, value2, value3]"

### DIFF
--- a/springwolf-add-ons/springwolf-generic-binding/src/main/java/io/github/springwolf/addons/generic_binding/annotation/processor/PropertiesUtil.java
+++ b/springwolf-add-ons/springwolf-generic-binding/src/main/java/io/github/springwolf/addons/generic_binding/annotation/processor/PropertiesUtil.java
@@ -16,7 +16,7 @@ import java.util.regex.Pattern;
 @Slf4j
 public class PropertiesUtil {
 
-    private static final Pattern ARRAY_PATTERN = Pattern.compile("\\[([^]]+)]");
+    private static final Pattern ARRAY_PATTERN = Pattern.compile("^\\[([^]]+)]$");
 
     private PropertiesUtil() {}
 
--- a/springwolf-add-ons/springwolf-generic-binding/src/main/java/io/github/springwolf/addons/generic_binding/annotation/processor/PropertiesUtil.java
+++ b/springwolf-add-ons/springwolf-generic-binding/src/main/java/io/github/springwolf/addons/generic_binding/annotation/processor/PropertiesUtil.java
@@ -16,7 +16,9 @@ import java.util.regex.Pattern;
 @Slf4j
 public class PropertiesUtil {
 
-    private static final Pattern ARRAY_PATTERN = Pattern.compile("\\[([^\\]]+)\\]");
+    private static final Pattern ARRAY_PATTERN = Pattern.compile("\\[([^]]+)]");
+
+    private PropertiesUtil() {}
 
     public static Map<String, Object> toMap(String[] propertyStrings) {
         return convertPropertiesToNestedMap(buildPropertiesFrom((propertyStrings)));
@@ -63,10 +65,7 @@ public class PropertiesUtil {
             if (matcher.find()) {
                 // Extract the key and values
                 String[] values = matcher.group(1).split(",");
-                for (int i = 0; i < values.length; i++) {
-                    values[i] = values[i].trim();
-                }
-                return Arrays.asList(values);
+                return Arrays.stream(values).map(String::trim).toList();
             }
         }
 
--- a/springwolf-add-ons/springwolf-generic-binding/src/main/java/io/github/springwolf/addons/generic_binding/annotation/processor/PropertiesUtil.java
+++ b/springwolf-add-ons/springwolf-generic-binding/src/main/java/io/github/springwolf/addons/generic_binding/annotation/processor/PropertiesUtil.java
@@ -10,9 +10,13 @@ import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.Map;
 import java.util.Properties;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 @Slf4j
 public class PropertiesUtil {
+
+    private static final Pattern ARRAY_PATTERN = Pattern.compile("\\[([^\\]]+)\\]");
 
     public static Map<String, Object> toMap(String[] propertyStrings) {
         return convertPropertiesToNestedMap(buildPropertiesFrom((propertyStrings)));
@@ -39,17 +43,33 @@ public class PropertiesUtil {
             Map<String, Object> mapNode = bindingData;
             while (path.size() > 1) {
                 String pathElement = path.get(0);
-                if (!mapNode.containsKey(pathElement)) {
-                    mapNode.put(pathElement, new HashMap<>());
-                }
-
+                mapNode.computeIfAbsent(pathElement, k -> new HashMap<>());
                 mapNode = (Map<String, Object>) mapNode.get(pathElement);
 
                 path.pop();
             }
 
-            mapNode.put(path.get(0), properties.get(propertyName));
+            var value = parseValue(properties.get(propertyName));
+            mapNode.put(path.get(0), value);
         }
         return bindingData;
+    }
+
+    private static Object parseValue(Object input) {
+        if (input instanceof String inputString) {
+            Matcher matcher = ARRAY_PATTERN.matcher(inputString);
+
+            // Check if the pattern matches the input string
+            if (matcher.find()) {
+                // Extract the key and values
+                String[] values = matcher.group(1).split(",");
+                for (int i = 0; i < values.length; i++) {
+                    values[i] = values[i].trim();
+                }
+                return Arrays.asList(values);
+            }
+        }
+
+        return input;
     }
 }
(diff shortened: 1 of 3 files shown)


springwolf-add-ons/springwolf-generic-binding/src/main/java/io/github/springwolf/addons/generic_binding/annotation/processor/PropertiesUtil.java
```diff
@@ -39,17 +43,33 @@ private static Map<String, Object> convertPropertiesToNestedMap(Properties prope
             Map<String, Object> mapNode = bindingData;
             while (path.size() > 1) {
                 String pathElement = path.get(0);
-                if (!mapNode.containsKey(pathElement)) {
-                    mapNode.put(pathElement, new HashMap<>());
-                }
-
+                mapNode.computeIfAbsent(pathElement, k -> new HashMap<>());
```
timonback, line 46: Nice refactor. Is the return value is identical to the get call in the next line? -> combine?

springwolf-add-ons/springwolf-generic-binding/src/main/java/io/github/springwolf/addons/generic_binding/annotation/processor/PropertiesUtil.java
```diff
@@ -10,10 +10,14 @@
 import java.util.LinkedList;
 import java.util.Map;
 import java.util.Properties;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 @Slf4j
 public class PropertiesUtil {
 
+    private static final Pattern ARRAY_PATTERN = Pattern.compile("\\[([^\\]]+)\\]");
```
timonback, line 19: I am not on a computer, so I cannot test it right now. Is it indented that the count of [ and ] do not match?

springwolf-add-ons/springwolf-generic-binding/src/main/java/io/github/springwolf/addons/generic_binding/annotation/processor/PropertiesUtil.java
```diff
@@ -39,17 +43,33 @@ private static Map<String, Object> convertPropertiesToNestedMap(Properties prope
...
+            Matcher matcher = ARRAY_PATTERN.matcher(inputString);
+
+            // Check if the pattern matches the input string
+            if (matcher.find()) {
+                // Extract the key and values
+                String[] values = matcher.group(1).split(",");
+                for (int i = 0; i < values.length; i++) {
+                    values[i] = values[i].trim();
+                }
+                return Arrays.asList(values);
```
timonback, line 69: Nitpicking: can this 4 line block be written using Java stream api?

springwolf-add-ons/springwolf-generic-binding/src/test/java/io/github/springwolf/addons/generic_binding/annotation/processor/AsyncGenericOperationBindingProcessorTest.java
```diff
@@ -136,5 +159,17 @@ void yamlSyntaxDoesWorkAsWell() {
             // then
             assertThat(result).isEqualTo(Map.of("key", "value"));
         }
+
+        @Test
+        void yamlSyntaxArrayPropertyTest() {
+            // given
+            String[] strings = {"key: [value1, value2, value3]"};
```
timonback, line 166: Comment: looks fine based on the yaml standard

I wonder whether the values should be escaped ['value1'], but that can be a future task/feature request

springwolf-add-ons/springwolf-generic-binding/src/test/java/io/github/springwolf/addons/generic_binding/annotation/processor/AsyncGenericOperationBindingProcessorTest.java
```diff
@@ -101,6 +100,30 @@ void twoPropertiesTest() {
             assertThat(result).isEqualTo(Map.of("key1", "value1", "key2", "value2"));
         }
 
+        @Test
+        void arrayPropertyTest() {
+            // given
+            String[] strings = {"key=[value1, value2, value3]"};
```
timonback, line 106: I read about the Java properties format and the native format is basically a key/value map of strings without support for further data types.
the closed about arrays/sequences that I found in the context of Spring is https://stackoverflow.com/a/33944154 which suggests the following format: ˋkey=value1, value2, value3ˋ (no square brackets)

Does this makes sense here?

springwolf-add-ons/springwolf-generic-binding/src/main/java/io/github/springwolf/addons/generic_binding/annotation/processor/PropertiesUtil.java
ctasada: It's exactly the same.

springwolf-add-ons/springwolf-generic-binding/src/main/java/io/github/springwolf/addons/generic_binding/annotation/processor/PropertiesUtil.java
ctasada: Good catch, it's even nicer 😄  `return Arrays.stream(values).map(String::trim).toList();`

springwolf-add-ons/springwolf-generic-binding/src/main/java/io/github/springwolf/addons/generic_binding/annotation/processor/PropertiesUtil.java
ctasada: That was the intention, but I found I can simplify it a bit more

springwolf-add-ons/springwolf-generic-binding/src/test/java/io/github/springwolf/addons/generic_binding/annotation/processor/AsyncGenericOperationBindingProcessorTest.java
ctasada: I saw it too, but it doesn't seem to work. Looking at java.util.Properties (which is what we use for the parsing) it doesn't seem to support arrays out-of-box

springwolf-add-ons/springwolf-generic-binding/src/test/java/io/github/springwolf/addons/generic_binding/annotation/processor/AsyncGenericOperationBindingProcessorTest.java
ctasada: A quick test shows that we can do `"key: [value1, value2, value3 is long]"` and it works fine. The moment we add `'` we need to provide extra escaping support.

We can do it, but I wonder if would make more sense to modify the annotation and provide something like
```
@AsyncGenericOperationBinding(
      type = "test-binding",
      yaml = """
                  something:
                     - 'long description'
                     - 'something else'
                  """
```
and use a YAML Parser for the field. That would provide complete support for any combination.

timonback: Totally fine. Just something I thought about. In case it becomes an issue, we can address it then

springwolf-add-ons/springwolf-generic-binding/src/test/java/io/github/springwolf/addons/generic_binding/annotation/processor/AsyncGenericOperationBindingProcessorTest.java
ctasada: My main concern is that removing the square brackets makes the parsing harder.

`{"array=1,2,3", "description=something, something else"}`

The first should be an array, while the second should be a string, but removing the `[]` it becomes more difficult to guess how should be parsed.

timonback: Good point! I guess the square brackets around the values is the most natural approach.

Will consult with @sam0r on Friday

springwolf-add-ons/springwolf-generic-binding/src/main/java/io/github/springwolf/addons/generic_binding/annotation/processor/PropertiesUtil.java
sam0r040: What do you think about this pattern `^\\[([^\\]]+)\\]$` because the square brackets are escaped and we explicitly match the beginning and the end of the string?
To avoid padded spaces at the beginning of the string, we could simply trim the string before applying the regex. 
We used this to verify the pattern:
```
[asdf,s]
[sd]
asdf[sdf]
[sdf][sdf]
[sd[sdf]]
[kdkd]dkkd
[kdkd
```

ctasada: @sam0r040 I just pushed a change adapting your proposal. With this approach only fully valid arrays are converted, anything else is kept as a String, which I think is a nice outcome.

timonback (approved):

Issue #665, opened 2024-03-21
KafkaListener method without any argument throw exception


alievrenkut:
**Describe the bug**
Springwolf throws exception at below if project has a method without any argument. 
`java.lang.IllegalArgumentException: Payload cannot be detected. Method must not have 0 parameters`

I think zero argument method should be ok and should supported by springwolf.

**Dependencies and versions used**
`springwolf-kafka` version `1.0.0`.

**Code example**
```java
    @KafkaListener(topics = KafkaTopics.Context.END_MISSION_TOPIC, containerFactory = "protobufKafkaListenerContainerFactory")
    public void endMission() {

    }
```

**Stack trace and error logs**
java.lang.IllegalArgumentException: Payload cannot be detected. Method must not have 0 parameters: MissionManagmentKafkaService::endMission
	at io.github.springwolf.core.asyncapi.scanners.common.payload.PayloadClassExtractor.getPayloadParameterIndex(PayloadClassExtractor.java:45)
	at io.github.springwolf.core.asyncapi.scanners.common.payload.PayloadClassExtractor.extractFrom(PayloadClassExtractor.java:33)
	at io.github.springwolf.core.asyncapi.scanners.operations.annotations.SpringAnnotationMethodLevelOperationsScanner.mapMethodToOperation(SpringAnnotationMethodLevelOperationsScanner.java:62)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:179)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:179)
	at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:1024)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:151)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:174)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:596)
[...]

PR #687, opened 2024-04-05
Feat/handle no payload

sam0r040:
resolves: GH-665

### DIFF
--- a/springwolf-add-ons/springwolf-common-model-converters/src/main/java/io/github/springwolf/addons/common_model_converters/converters/monetaryamount/MonetaryAmountConverter.java
+++ b/springwolf-add-ons/springwolf-common-model-converters/src/main/java/io/github/springwolf/addons/common_model_converters/converters/monetaryamount/MonetaryAmountConverter.java
@@ -14,18 +14,32 @@ public class MonetaryAmountConverter implements ModelConverter {
 
     @Override
     public Schema resolve(AnnotatedType type, ModelConverterContext context, Iterator<ModelConverter> chain) {
-        JavaType javaType = Json.mapper().constructType(type.getType());
-        if (javaType != null) {
-            Class<?> cls = javaType.getRawClass();
-            if (javax.money.MonetaryAmount.class.isAssignableFrom(cls)) {
-                type = new AnnotatedType(MonetaryAmount.class).resolveAsRef(true);
-                Schema<?> schema = (chain.hasNext()) ? chain.next().resolve(type, context, chain) : null;
-                if (schema != null) {
-                    schema.name(javaType.getRawClass().getName());
-                }
-                return schema;
-            }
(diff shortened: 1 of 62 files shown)


springwolf-add-ons/springwolf-common-model-converters/src/main/java/io/github/springwolf/addons/common_model_converters/converters/monetaryamount/MonetaryAmountConverter.java
```diff
@@ -19,7 +19,11 @@ public Schema resolve(AnnotatedType type, ModelConverterContext context, Iterato
             Class<?> cls = javaType.getRawClass();
             if (javax.money.MonetaryAmount.class.isAssignableFrom(cls)) {
                 type = new AnnotatedType(MonetaryAmount.class).resolveAsRef(true);
-                return (chain.hasNext()) ? chain.next().resolve(type, context, chain).name(javaType.getRawClass().getName()) : null;
+                Schema<?> schema = (chain.hasNext()) ? chain.next().resolve(type, context, chain) : null;
```
sam0r040, line 22: Should we have just one call for chaining in the code? I am a bit worried that the second call might be overlooked in the future and cause problems. On the other hand it is clear when something is changed and when not because all the manipulation is in one place.

springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/components/DefaultComponentsService.java
```diff
@@ -145,10 +145,25 @@ private String getSchemaName(Class<?> type, Map<String, Schema> schemas) {
...
 
+    private void processCommonModelConverters(Map<String, Schema> schemas) {
+        schemas.values().stream()
+                .filter(schema -> schema.getType() == null)
+                .filter(schema -> schema.get$ref() != null)
+                .forEach(schema -> {
+                    String targetSchemaName = schema.getName();
+                    String sourceSchemaName = StringUtils.substringAfterLast(schema.get$ref(), "/");
+
+                    Schema<?> actualSchema = schemas.get(sourceSchemaName);
```
sam0r040, line 161: Line 159 and 161 seem to belong together as the are actually resolving a schema from a ref. I am wondering why you split them up?

springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/components/DefaultComponentsService.java
```diff
@@ -145,10 +145,25 @@ private String getSchemaName(Class<?> type, Map<String, Schema> schemas) {
...
+                    schemas.put(targetSchemaName, actualSchema);
```
sam0r040, line 162: We might put nulls in into the map, is it intentional?

PR #691, opened 2024-04-07
fix: Fixes description indenting

ctasada:
When the description field includes a long description it may not be properly rendered, mainly in Kotlin, where the way of managing the text-blocks is different.

 This PR improves the way descriptions are managed by:

 - Removes empty lines at the beginning of the text-block
 - Removes indention due to alignment, but keeps indention in the lines that need it

### DIFF
--- a/springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/scanners/common/utils/DescriptionUtil.java
+++ b/springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/scanners/common/utils/DescriptionUtil.java
@@ -1,0 +1,74 @@
+// SPDX-License-Identifier: Apache-2.0
+package io.github.springwolf.core.asyncapi.scanners.common.utils;
+
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.Arrays;
+
+public class DescriptionUtil {
+    private DescriptionUtil() {}
+
+    /**
+     * Returns a string where all leading indentation has been removed from each line.
+     * It detects the smallest common indentation of all the lines in the input string,
+     * and removes it.
+     *
+     * @param text The original string with possible leading indentation.
+     * @return The string with leading indentation removed from each line.
+     */
+    public static String trimIndent(String text) {
+        if (text == null) {
+            return null;
+        }
+        final String newLine = System.lineSeparator();
+        String[] lines = text.split(newLine);
+        int firstNonEmptyLine = findFirstNonEmptyIndex(lines);
+        if (firstNonEmptyLine != -1) {
+            lines = Arrays.copyOfRange(lines, firstNonEmptyLine, lines.length);
+        }
+        int minIndent = resolveMinIndent(lines);
+        var result = Arrays.stream(lines)
+                .map(line -> line.substring(Math.min(line.length(), minIndent)))
+                .reduce((a, b) -> a + newLine + b)
+                .orElse(StringUtils.EMPTY);
+
+        if (endsWithNewline(text)) {
+            result = result.concat(newLine);
+        }
+
+        return result;
+    }
+
+    private static int findFirstNonEmptyIndex(String[] lines) {
+        for (int i = 0; i < lines.length; i++) {
+            if (!lines[i].trim().isEmpty()) {
+                return i; // Return index of first non-empty line
+            }
+        }
+        return -1; // If all lines are empty, return -1
+    }
+
+    private static boolean endsWithNewline(String text) {
+        if (text == null || text.isEmpty()) {
+            return false;
+        }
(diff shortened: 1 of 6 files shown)


ctasada:
I think this issue may be exclusive from Kotlin, when using text-blocks. The change should be transparent (or nearly transparent) for Java code, but we can wrap it in a special setting if needed.

timonback:
I am assuming you use multi–line strings aka text blocks.
If it is kotlin specific, would it be sufficient for the users to call trimIndent themselves?

https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/trim-indent.html

(I haven't looked through the code. This sounds like a method that should be part of a standard lib)

ctasada:
Sadly it's not possible. Annotations must be compile-time constants, so it's not possible to do any manipulation, like `.trimIndent()`

timonback:
Argh. Unfortunate that https://youtrack.jetbrains.com/issue/KT-22505/Support-calls-to-trimMargin-trimIndent-as-initializer-for-const-vals is not getting resolved.
And there seems to be no easy Apache Commons or similar implementation.

Will review the PR on Friday

springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/scanners/common/utils/DescriptionUtil.java
```diff
@@ -0,0 +1,83 @@
...
+    private static int resolveMinIndent(String[] lines) {
+        return Arrays.stream(lines)
+                .filter(line -> !line.trim().isEmpty())
+                .mapToInt(DescriptionUtil::countLeadingSpaces)
+                .min()
+                .orElse(0);
+    }
+
+    private static int countLeadingSpaces(String line) {
+        for (int i = 0; i < line.length(); i++) {
```
sam0r040, line 76: I think we could use StringUtils.indexOfAnyBut(line, ' ') instead

springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/scanners/common/utils/DescriptionUtil.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * @return {@code true} if the line ends with a {@code '\n'} or {@code false} otherwise
+     */
+    private static boolean endsWithNewline(String text) {
+        if (text == null || text.isEmpty()) {
+            return false;
+        }
+        return text.charAt(text.length() - 1) == '\n';
```
sam0r040, line 64: We should be able to use `StringUtils.endsWith(text, "\n")` instead

springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/scanners/common/utils/DescriptionUtil.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+        final String newLine = System.lineSeparator();
+        String[] lines = text.split(newLine);
+        int firstNonEmptyLine = findFirstNonEmptyIndex(lines);
+        if (firstNonEmptyLine != -1) {
+            lines = Arrays.copyOfRange(lines, firstNonEmptyLine, lines.length);
+        }
+        int minIndent = resolveMinIndent(lines);
+        var result = Arrays.stream(lines)
+                .map(line -> line.substring(Math.min(line.length(), minIndent)))
```
sam0r040, line 34: Could we simplify by using `StringUtils.substring(line, minIndent)`?

sam0r040:
Kind of unfortunate that there is no nice solution and we have to provide the Utils for it. I think I found some simplifications, but not as much as I wished :(. However, I would add two more test cases: String with no empty lines and null input.

springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/scanners/common/utils/DescriptionUtil.java
```diff
@@ -0,0 +1,83 @@
...
+package io.github.springwolf.core.asyncapi.scanners.common.utils;
+
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.Arrays;
+
+/**
+ * Auxiliary class to manage Description or long texts.
+ */
+public class DescriptionUtil {
```
sam0r040, line 11: Have you considered naming it TextUtil? I think the functionality is independent of descriptions.

ctasada:
@sam0r040 All the comments addressed. I also added more tests covering the different corner cases. Coverage is now 100%

timonback (approved):

Issue #700, opened 2024-04-14
@KafkaAsyncOperationBinding "type" doesn't work in case of using it with @KafkaHandler


stammaja:
In a setup similar to your example ExampleClassLevelKafkaListener, I tried to use KafkaAsyncOperationBinding "type" in order to separate  listeners per server they listen to. But I'm getting "kafka" instead.

`springwolf-kafka` version `1.1.0`.

@KafkaListener(topics = "${push-api.topic}", groupId = "${push-api.topic.consumer-group}")
public class PushApiListener {

    @KafkaHandler
    @KafkaAsyncOperationBinding(
            type = "kafka-external"
    )
    public void handleNotification(NotificationDto notification) {
        ...
    }

springwolf.docket.servers.kafka-external.protocol=kafka-secure
springwolf.docket.servers.kafka-external.host=...

Expected: In the UI listener will be shown with tag "kafka-external"
Actual: It is shown still as "kafka"

timonback:
To better indicate in the ui the difference between protocol and server name, https://github.com/springwolf/springwolf-core/pull/738 updates the examples

Help is still welcome to show the server name as part of the channel item

timonback:
On this is being worked in https://github.com/springwolf/springwolf-core/pull/832

PR #709, opened 2024-04-19
chore(bindings)!: remove type field from protocol specific binding annotations

sam0r040:
the field is fixed by the spec and should not be modified by the user


timonback:
Relates to https://github.com/springwolf/springwolf-core/issues/700

timonback (approved):

PR #713, opened 2024-04-20
(feat) Polymorphism support for kotlinx serialization

### LINKED ISSUE #707: Polymorphism support for kotlinx.serialization
Given a polymorphic hierarchy like this:

```kotlin
@Serializable
sealed interface Event

@Serializable
@SerialName("PostAdded")
data class PostAdded(val message: String) : Event

@Serializable
@SerialName("PostDeleted")
data class PostDeleted(val id: Int): Event
```

I would expect the generated schema to contain a `oneOf: PostAdded, PostDeleted` for `Event`, along with a discriminator property, such as it does for Jackson.

**Motivation**
We have several events that used `sealed interface` and polymorphism. This maps cleanly to `oneOf` in OpenAPI and AsyncAPI, and should be supported.

ctasada:
Fixes #707. Adds basic polymorphism support to Kotlinx serialization

### DIFF
--- a/springwolf-add-ons/springwolf-kotlinx-serialization-model-converter/src/main/kotlin/io/github/springwolf/addons/kotlinx_serialization_model_converter/converter/KotlinxSerializationModelConverter.kt
+++ b/springwolf-add-ons/springwolf-kotlinx-serialization-model-converter/src/main/kotlin/io/github/springwolf/addons/kotlinx_serialization_model_converter/converter/KotlinxSerializationModelConverter.kt
@@ -6,6 +6,7 @@ import io.swagger.v3.core.converter.ModelConverter
 import io.swagger.v3.core.converter.ModelConverterContext
 import io.swagger.v3.core.util.RefUtils
 import io.swagger.v3.oas.models.media.ArraySchema
+import io.swagger.v3.oas.models.media.ComposedSchema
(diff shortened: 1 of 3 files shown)


springwolf-add-ons/springwolf-kotlinx-serialization-model-converter/src/main/kotlin/io/github/springwolf/addons/kotlinx_serialization_model_converter/converter/KotlinxSerializationModelConverter.kt
```diff
@@ -60,12 +61,18 @@ class KotlinxSerializationModelConverter(private val useFqn: Boolean = false) :
 
     private fun getPropertySchema(property: KProperty1<*, *>, context: ModelConverterContext): Schema<*> {
         val propertySchema: Schema<*>
-
-       val name = getPropertyName(property)
-
+        val name = getPropertyName(property)
```
timonback, line 64: Seems like spotless is not configured to handle kotlin files as well

Could be a great future addition in https://github.com/springwolf/springwolf-core/blob/a3198a4cfbabe852c2ab425094ef478c649ab598/build.gradle#L54

timonback (approved):
Looks great and actually quite straight forward!

PR #722, opened 2024-04-26
fix(core): format all Date, LocalDate and OffsetDateTimes in iso format

sam0r040:


springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/components/examples/walkers/DefaultSchemaWalker.java
```diff
@@ -123,6 +131,16 @@ private Optional<T> getExampleValueFromSchemaAnnotation(Schema schema) {
             return exampleValueGenerator.createDoubleExample(doubleValue, schema);
         }
 
+        if (exampleValue instanceof Date exampleDate) {
```
timonback, line 134:
```suggestion
        if (exampleValue instanceof Date exampleDate) { // in case of LocalDate, swagger-parser converts it into a Date
```

Issue #730, opened 2024-04-28
JSON object with inheritance is not rendering the parent object correctly


fcmdeveloper1:
Dears,
    i am not sure if this is the behavior of the library or not, but i can see that if i have an object that inherits from a parent object, not all the parent object fields are generated in the specs, specially if there is a "discriminatorProperty".

also, the above is not rendered unless i included both the discriminatorProperty = "" and allOf = {ParentEventDto.class} in the @schema field of the child object, and even if i did this inclusion, the UI is not rendered correctly for the child object.(missing a lot of fields)

for swagger, without adding the allOf and discriminatorProperty attributes, everything is working perfectly, which is the expected behavior for both UI and the generated swagger description.

![image](https://github.com/springwolf/springwolf-core/assets/148101792/ac51401c-6ca2-4383-b542-6a4f5259ee9e)

disclaimer, i am using the exact same objects with the same @schema attributes for both description swagger and Kafkaconsumer

My Best Regards

timonback:
Hi @fcmdeveloper1 ,

Can you share the classes that are involved, again so that we can easily reproduce it?
Or are those the same as in https://github.com/springwolf/springwolf-core/issues/729 ?

An alternative to uploading the code as zip is to fork the repo, create a new branch, edit the example (https://github.com/springwolf/springwolf-core/blob/master/springwolf-examples/springwolf-kafka-example/src/main/java/io/github/springwolf/examples/kafka/dtos/ExamplePayloadDto.java ; no need to fix tests or anything like that) and commit&push + open a pr to the repo (or your fork)

fcmdeveloper1:
Dear @timonback,
       yes, exact project, you can reproduce the issue, also, you can see the behavior if you omit the "allOf" and "discriminatorProperty" from the LearningEvent object which is the desired setup to match the swagger default behavior, the eventKey property will not appear in the object.

please, let me know if i can help or support.

My Best Regards

timonback:
Hi @fcmdeveloper1,

I started looking into and created https://github.com/springwolf/springwolf-core/pull/745 so that we can look at the same thing together.

The code contains two variations - see the commits in the pr
1. The discriminator field `eventKey` is added as own field in the Parent class
2. `LearningEvent` has the `@Schema#allOf` property set

- Variation 1 works out of the box for me, although requires code changes in your event class.
[...]

PR #731, opened 2024-04-28
(feat) Add support to parse Annotations from super classes

ctasada:
When a class is inheriting annotated methods from a superclass, the code is now properly parsing those methods for annotations.

This development allows to parse annotations from abstract classes or other inheritance classes, where the annotations are defined in a common class.

### DIFF
--- a/springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/scanners/common/AsyncAnnotationScanner.java
+++ b/springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/scanners/common/AsyncAnnotationScanner.java
@@ -26,6 +26,7 @@ import io.github.springwolf.core.asyncapi.scanners.common.utils.TextUtils;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.context.EmbeddedValueResolverAware;
+import org.springframework.util.ReflectionUtils;
 import org.springframework.util.StringUtils;
 import org.springframework.util.StringValueResolver;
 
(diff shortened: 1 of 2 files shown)


springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/scanners/common/AsyncAnnotationScanner.java
```diff
@@ -58,14 +58,27 @@ protected Stream<MethodAndAnnotation<A>> getAnnotatedMethods(Class<?> type) {
...
                 .flatMap(method -> AnnotationScannerUtil.findAnnotations(annotationClass, method).stream()
                         .map(annotation -> new MethodAndAnnotation<>(method, annotation)));
     }
 
+    /**
+     * Recursively gets the list of declared methods of the class' parent classes
+     * @param type Class to retrieve the list of declared methods
+     * @return List of declared methods
+     */
+    private List<Method> getInheritedMethods(Class<?> type) {
```
sam0r040, line 74: Could we use org.springframework.util.ReflectionUtils#getAllDeclaredMethods for this?

sam0r040:
Hi @ctasada is this related to https://github.com/springwolf/springwolf-core/issues/730 ?

ctasada:
Hey @sam0r040 , no, this PR was not intended to fix #730. The issue I'm fixing is when we have the `@AsyncListener` defined in an `abstract` class.

sam0r040:
Ahh okay then we need to investigate separately.

timonback (approved):

PR #733, opened 2024-04-29
(fix) Fixed a bug loading the Kotlinx Module without `use-fqn` variable

ctasada:
When the Kotlinx Serialization Module was available in an environment without the `springwolf.use-fqn` the code was failing, since it was trying to inject the variable.

The code is now loading the converter **only** if `springwolf.enabled` is `true` (default value). The `springwolf.use-fqn` is also defaulting to `true`.

This solution fixes the identified issue.

### DIFF
--- a/springwolf-add-ons/springwolf-kotlinx-serialization-model-converter/src/main/java/io/github/springwolf/addons/kotlinx_serialization_model_converter/configuration/KotlinxSerializationModelConverterAutoConfiguration.java
+++ b/springwolf-add-ons/springwolf-kotlinx-serialization-model-converter/src/main/java/io/github/springwolf/addons/kotlinx_serialization_model_converter/configuration/KotlinxSerializationModelConverterAutoConfiguration.java
@@ -3,12 +3,10 @@ package io.github.springwolf.addons.kotlinx_serialization_model_converter.config
 
 import io.github.springwolf.addons.kotlinx_serialization_model_converter.converter.KotlinxSerializationModelConverter;
 import io.github.springwolf.core.configuration.properties.SpringwolfConfigConstants;
-import org.springframework.beans.factory.annotation.Value;
+import io.github.springwolf.core.configuration.properties.SpringwolfConfigProperties;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
-
-import static io.github.springwolf.core.configuration.properties.SpringwolfConfigConstants.SPRINGWOLF_USE_FQN;
 
 /**
  * Spring AutoConfiguration adding an {@link KotlinxSerializationModelConverter} Bean to the spring context.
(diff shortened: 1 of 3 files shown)


springwolf-add-ons/springwolf-kotlinx-serialization-model-converter/src/main/java/io/github/springwolf/addons/kotlinx_serialization_model_converter/configuration/KotlinxSerializationModelConverterAutoConfiguration.java
```diff
@@ -3,18 +3,20 @@
...
  * Spring AutoConfiguration adding an {@link KotlinxSerializationModelConverter} Bean to the spring context.
  */
 @Configuration(proxyBeanMethods = false)
+@ConditionalOnProperty(name = "springwolf.enabled", havingValue = "true", matchIfMissing = true)
 public class KotlinxSerializationModelConverterAutoConfiguration {
 
     @Bean
     public KotlinxSerializationModelConverter kotlinxSerializationTypeConverter(
-            @Value("${springwolf.use-fqn}") boolean useFqn) {
+            @Value("${springwolf.use-fqn:true}") boolean useFqn) {
```
sam0r040, line 19: I was not aware that we have an implicit dependency to core configuration here. For me SpringwolfConfigProperties is the source of truth for configuration. Therefore I would suggest to add the dependency to core and use the the properties object.

ctasada: I thought about that, but I was hesitant to couple both modules. I will do the changes later today.

timonback (approved):

springwolf-add-ons/springwolf-kotlinx-serialization-model-converter/src/main/java/io/github/springwolf/addons/kotlinx_serialization_model_converter/configuration/KotlinxSerializationModelConverterAutoConfiguration.java
```diff
@@ -2,19 +2,24 @@
...
  * Spring AutoConfiguration adding an {@link KotlinxSerializationModelConverter} Bean to the spring context.
  */
 @Configuration(proxyBeanMethods = false)
+@ConditionalOnProperty(name = SpringwolfConfigConstants.SPRINGWOLF_ENABLED, havingValue = "true", matchIfMissing = true)
 public class KotlinxSerializationModelConverterAutoConfiguration {
 
     @Bean
     public KotlinxSerializationModelConverter kotlinxSerializationTypeConverter(
-            @Value("${springwolf.use-fqn}") boolean useFqn) {
+            @Value("${" + SPRINGWOLF_USE_FQN + ":true}") boolean useFqn) {
```
timonback, line 22: Are you able to inject the complete Springwolf ConfigDocket, which has the useFqn property? 
In contains the correct default value as well

timonback (approved):

Issue #609, opened 2024-02-19
Feature to hide buttons inside kafka publish operations


achabdo:
**Possibilité to hide buttons inside kafka publish or subscribe operation**
We want an option to hide unnecessary buttons that can be shown to users using our Kafka documentation.

**Motivation**
The problem is that users can send Kafka notifications to our Kafka topics, which can be dangerous. When you send a Kafka message to our broker servers, this data is used as replication, and it 's inserted into the database. We don't want users to have the ability to do that, and we want to hide completely the publish button, as shown in the following image.
![Capture](https://github.com/springwolf/springwolf-core/assets/11176293/20c060d8-6e1c-4035-87b2-d61a03790fc1)

**Technical details**
We want to have the ability to configure that by using  AsyncApiDocket asyncApiDocket() even if it 's a deprecated methode as we still using it in our project. Or by using the application.yml in the future.
For exemple :
var docket =  AsyncApiDocket.builder().hidePublishButton(true)

timonback:
Hi @achabdo , true, the ui shows all functionality regardless of how Springwolf is configured.

To enable publishing, you have to explicitly activate the publishing setting in Springwolf (i.e. `springwolf.plugin.kafka.publishing.enabled=true`).

Just to clarify, you want to hide the publish button in the ui, when the publishing functionality is disabled in Springwolf?

achabdo:
Yes , I want to hide the publish button when the publishing is disabled as users may think that they have the ability to send kafka notifications which can lead to confusion.

timonback:
Got it.

For this, Springwolf-ui would need to get the information from the backend. The backend would need to offer the information through some kind of new endpoint — combining the information from each activated plugin.

I see also another solution, which is easy to implement and might be good enough. The button stays visible and when pressed, a better error message is shown (i.e.: publishing failed. This feature might have been disabled by the administrator of this application). Based on the http response status code, the authorization error can differentiated.

Is the second option good enough for you?
Feel free to get started on it.

achabdo:
I will go with the first option. Thank you.

PR #766, opened 2024-05-20
feat(core): show publish button in ui only when publishing is enabled

timonback:
This uses a trivial approach of adding a GET endpoint per protocol (no single combined one) to indicate to the ui, whether publishing is enabled.

Relates to https://github.com/springwolf/springwolf-core/issues/609

### DIFF
--- a/springwolf-ui/src/app/components/channels/channel-main/channel-main.component.ts
+++ b/springwolf-ui/src/app/components/channels/channel-main/channel-main.component.ts
@@ -32,7 +32,6 @@ export class ChannelMainComponent implements OnInit {
   messageBindingExample?: Example;
   messageBindingExampleTextAreaLineCount: number = 1;
   canPublish: boolean = false;
-  canPublishSubscription?: Subscription;
 
   constructor(
     private asyncApiService: AsyncApiService,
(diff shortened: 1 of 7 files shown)


springwolf-ui/src/app/components/channels/channel-main/channel-main.component.ts
```diff
@@ -57,6 +60,15 @@ export class ChannelMainComponent implements OnInit {
       this.headersTextAreaLineCount = this.headersExample?.lineCount || 1;
       this.messageBindingExampleTextAreaLineCount =
         this.messageBindingExample?.lineCount || 1;
+
+      if (this.canPublishSubscription !== undefined) {
```
sam0r040, line 64: Why do we need to unsubscribe?

timonback: Because I dont know Angular/rxjs well enough.
You are right, http get is a finite value and therefore does not needs this
Source: https://stackoverflow.com/a/41177163

springwolf-ui/src/app/components/channels/channel-main/channel-main.component.ts
```diff
@@ -57,6 +60,15 @@ export class ChannelMainComponent implements OnInit {
...
+        this.canPublishSubscription.unsubscribe();
+      }
+      this.canPublishSubscription = this.publisherService
+        .canPublish(this.operation.protocol)
+        .subscribe((response) => {
```
sam0r040, line 65: Maybe wee need to unsubscribe on ngOnDestroy?

PR #787, opened 2024-06-01
Read spring header annotation

timonback:
This uses https://github.com/springwolf/springwolf-core/pull/579 as a starting point and relates to https://github.com/springwolf/springwolf-core/issues/563

With this change:
- Springwolf is able to parse spring `@Header` annotations for spring method and class listeners like `@KafkaListener` (`@Headers` cannot be parsed, as it is a map that is populated during runtime method invocation)
- The Kafka Key is mapped into the message binding (additionally), the key is kept in the header as well, which is demonstrated in the example
- The `ComponentsService` becomes a pure stateful services that only stores swagger schema data. The converting of class to a swagger schema is moved to the new `SwaggerSchemaService` (refactor)
- The AsyncHeader signature gets an additional parameter headerSchema to pass the info for Kafka. I am not 100% sure about this change, but it seems easiest at this point.

CC: @ctasada


springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/schemas/SwaggerSchemaService.java
```diff
@@ -0,0 +1,213 @@
...
+            SwaggerSchemaUtil swaggerSchemaUtil,
+            SpringwolfConfigProperties properties) {
+
+        externalModelConverters.forEach(converter::addConverter);
+        this.schemaPostProcessors = schemaPostProcessors;
+        this.swaggerSchemaUtil = swaggerSchemaUtil;
+        this.properties = properties;
+    }
+
+    public record NewSchema(String schemaName, Map<String, Schema> schemas) {
```
timonback, line 50: NamedSwaggerSchema

PR #790, opened 2024-06-05
Units test to investigate AMQP generation 

pdalfarr:
Units test to investigate AMQP generation
I added some comments in the PR.

Issue #366, opened 2023-09-25
In Springwolf UI, add support for sending RabbitMQ Message towards an exchange using a RoutingKey


pdalfarr:
**Describe the feature request**
In Springwolf UI, add support for sending RabbitMQ Message towards an exchange using a Routing Key.

**Motivation**
Springwolf is able to list all the Publishers and Subscribers and present a UI which, not only list these, but also allow to test them by sending message.
Regarding RabbitMq, this 'message sending' feature is working fine for RabbitMQ QUEUES, but does not work for EXCHANGE.
This feature request aims to implements this missing part.

**Technical details**

Here are some technical details I gathered so far:

The springwolf UI allows to send messages to QUEUES (when plugin.amqp.publishing.enabled: true )
which is OK when specifying queue name in AsyncOperation.channelName

But sending a message to a TOPIC (exchange) is not working as expected:

Let's take an example:

In one of the RabbitMQ your example, here

https://github.com/springwolf/springwolf-core/blob/master/springwolf-examples/springwolf-amqp-example/src/main/java/io/github/stavshamir/springwolf/example/amqp/producers/ExampleProducer.java#L17-L28

we have:

```
   @AsyncPublisher(
            operation =
                    @AsyncOperation(
                            channelName = "example-producer-channel-publisher",
                            description = "Custom, optional description defined in the AsyncPublisher annotation"))
    @AmqpAsyncOperationBinding()
    public void sendMessage(ExamplePayloadDto msg) {
        // send
        AnotherPayloadDto dto = new AnotherPayloadDto("fooValue", msg);
        rabbitTemplate.convertAndSend("example-topic-exchange", "example-topic-routing-key", dto);
    }
```

Now, let's suppose we have, in another micro-service, a method with the following annotation:

```
    @Override
    @RabbitListener(
            bindings = @QueueBinding(
                    value = @Queue(
                            name = "microservice-2-private-queue-name" // this is a private queue of this micro-service: I do not want this string value to be known by the first micro-service (micro-service 1), i.e., your sample code
                    ),
                    exchange = @Exchange(
                            name = "example-producer-channel-publisher", // the exchange from in your sample code
                            type = ExchangeTypes.TOPIC
                    ),
                    key = "example-topic-routing-key" // the key from your sample code
            )
    )
```

So, this microservice2 does declare a topic which receives messages send by the RabbitMQ 'sample' code :

[...]

PR #886, opened 2024-07-31
fix: amqp channel name when using exchange

timonback:
rebased and continued version of https://github.com/springwolf/springwolf-core/pull/790
relates to https://github.com/springwolf/springwolf-core/issues/366


springwolf-examples/springwolf-amqp-example/src/test/resources/asyncapi.json
```diff
@@ -119,6 +140,69 @@
...
+            "durable": true,
+            "exclusive": false,
+            "autoDelete": false,
+            "vhost": "/"
+          },
+          "bindingVersion": "0.3.0"
+        }
+      }
+    },
+    "queue-read_#_CRUD-topic-exchange-2": {
```
timonback, line 165: Does this make sense?

This is an entry of type `is=routingKey`, which belongs to an exchange.
However the routingKey is not specified.
Instead, the routingKey is part of the channelIdentifier - which has no special meaning.

Does the spec require the field routingKey?
Similarly, is the target queue (binding) needed as well, especially since the routingKey can also be a pattern.

springwolf-plugins/springwolf-amqp-plugin/src/main/java/io/github/springwolf/plugins/amqp/asyncapi/scanners/bindings/RabbitListenerUtil.java
```diff
@@ -196,6 +225,7 @@ public static Map<String, OperationBinding> buildOperationBinding(
         return Map.of(
                 "amqp",
                 AMQPOperationBinding.builder()
+                        // TODO: cc for publishing does not match listener -> remove?
```
timonback, line 228: TODO: review

springwolf-plugins/springwolf-amqp-plugin/src/main/java/io/github/springwolf/plugins/amqp/asyncapi/scanners/bindings/RabbitListenerUtil.java
```diff
@@ -27,22 +27,35 @@
...
+ * </ul>
+ * <br/>
+ * How does rabbitmq work?
+ * <ol>
+ * <li> Producer sends a message to an exchange (default exchange if not specified)
+ * <li> Exchange routes the message to a queue based on the routing key (routing key = queue name if not specified)
+ * <li> Consumer consumes the message from the queue
+ * </ol>
+ */
+// TODO: should this do validation and throw errors when an invalid rabbit configuration is found?
```
timonback, line 47: TODO

Issue #829, opened 2024-07-02
Build example for schema: error with SpringWolf 1.4.0, but working fine with 1.2.0


pdalfarr:
**Describe the bug**
Schema generation was working fine with SpringWolf 1.2.0.
I changed version from 1.2.0 to 1.4.0: as a result, my Spring app is working fine, but Springwolf does not work.
See log message and stack trace below.

**Dependencies and versions used**
SpringWolf 1.2.0 :

		
		<dependency>
			<groupId>io.github.springwolf</groupId>
			<artifactId>springwolf-core</artifactId>
			<version>${version.springwolf}</version>
		</dependency>
		<dependency>
			<groupId>io.github.springwolf</groupId>
			<artifactId>springwolf-amqp</artifactId>
			<version>${version.springwolf}</version>
		</dependency>
		<dependency>
			<groupId>io.github.springwolf</groupId>
			<artifactId>springwolf-ui</artifactId>
			<version>${version.springwolf}</version>
		</dependency>
		

SpringWolf 1.4.0: same as above plus:

```
		<dependency>
			<groupId>io.github.springwolf</groupId>
			<artifactId>springwolf-amqp-binding</artifactId>
			<version>${version.springwolf}</version>
		</dependency>
```

**Code example**
No code example

**Stack trace and error logs**

I see multiple log messages like this :

```
2024-07-02 10:15:48.297  INFO [rapport-controle,,] 50188 --- [           main] i.g.s.c.a.c.e.w.DefaultSchemaWalker      : Failed to build example for schema: There is no name set for Schema: class StringSchema {
    class Schema {
        type: string
        format: null
        $ref: null
        description: null
        title: null
        multipleOf: null
        maximum: null
        exclusiveMaximum: null
        minimum: null
        exclusiveMinimum: null
        maxLength: null
        minLength: null
        pattern: null
        maxItems: null
        minItems: null
        uniqueItems: null
        maxProperties: null
        minProperties: null
        required: null
        not: null
        properties: null
        additionalProperties: null
        nullable: null
        readOnly: null
        writeOnly: null
        example: null
        externalDocs: null
        deprecated: null
        discriminator: null
        xml: null
    }
}

io.github.springwolf.core.asyncapi.components.examples.walkers.SchemaWalker$ExampleGeneratingException: There is no name set for Schema: class StringSchema {
    class Schema {
        type: string
        format: null
        $ref: null
        description: null
        title: null
        multipleOf: null
        maximum: null
[...]

PR #888, opened 2024-08-01
Feat/handle enum as ref schemas

timonback:
Closes https://github.com/springwolf/springwolf-core/pull/887
Relates to https://github.com/springwolf/springwolf-core/issues/829
Requires https://github.com/springwolf/springwolf-core/pull/875 to be merged first


springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/schemas/SwaggerSchemaService.java
```diff
@@ -184,6 +185,14 @@ private void processAsyncApiPayloadAnnotation(Map<String, Schema> schemas, Strin
         }
     }
 
+    private void processSchemaWithoutName(Map<String, Schema> schemas) {
+        schemas.forEach((schemaName, schema) -> {
+            if (schema.getName() == null) {
+                schema.setName(schemaName);
```
timonback, line 191: I am wondering, whether this is the better fix, since the Map from swagger contains the correct class name, but it is not set in the schema#name

At the same time, I am unsure, whether we should touch/modify the schema for this?

PR #923, opened 2024-08-16
feat(core): add SchemaTitleModelConverter

timonback:
extracted from https://github.com/springwolf/springwolf-core/pull/890


timonback:
@dabeck81 Is it expected that the `SchemaTitleModelConverter` does not change any tests assertions?

dabeck81:
> @dabeck81 Is it expected that the `SchemaTitleModelConverter` does not change any tests assertions?

As it is a spring bean, it will only affect tests that load the spring context. Tests that do not load spring context will remain asis.

timonback:
Fair enough. The example ApiIntegrationTest make use of the full spring context and their AsyncAPI.json did not change. Maybe they are missing a scenario that you covered in your test. Looking forward to your updated PR.

You probably noticed already, today I extracted and merged multiple aspects of your PR to simplify future reviews.

Issue #1071, opened 2024-11-12
Springwolf spring-cloud-stream should support BiConsumer


LeovR:
**Describe the feature request**
With Spring Cloud Stream one is able, in addition to `Function` and `Consumer`-Beans to use `BiFunction` or `BiConsumer`-Beans.
E.g. with a Kafka-Binder, a `BiConsumer` can be used to also consume the Kafka-Headers.
Currently, only `Consumer`, `Function` or `Supplier` can be used.

In addition, the `DefaultBeanMethodsScanner` should also consider all `@Component`-Beans and not just bean-methods in `@Configuration`-classes because those can also be provided via other `@Component`-annotations.

**Motivation**
We're relying on `BiConsumer` to consume headers.

**Technical details**
It should be considered, whether the Spring `FunctionCatalog` could be used to get a hold of the function-bindings, since all the scanned functions are already registered in the catalog.

timonback:
Hello @LeovR ,
that is a great addition.

Do you have a code example for the usage with headers? (Can be added to the cloudstream example later)
In the docs, I found a section where messages from two different topic are consumed: https://docs.spring.io/spring-cloud-stream/reference/kafka/kafka-streams-binder/programming-model.html#multiple-input-bindings

LeovR:
Hi @timonback,

unfortunately this feature is not documented very well.
The link you provided is about the Kafka Streams binding and not the "normal" Kafka binding.

I will pick this up myself then and create a pull request when I'm done.

At least I wanted to make sure whether this is wanted.

timonback:
Thank you @LeovR for adding the functionality of `BiConsumer`, it is already visible on the demo website [0] and about to be part of the `-SNAPSHOT` artifacts.

Do you plan to continue on i.e. BiFunction, FunctionCatalog, native cloudstream publishing?
If not, I will mark this issue as completed (you can always re-open/create a new one)

[0] https://cloud-stream.demo.springwolf.dev/springwolf-ui/asyncapi-ui.html#channel-biconsumer-topic

LeovR:
@timonback 
Maybe we should close this one and I will create new issues for the other parts.
I don't know yet when I will be able to do those.

PR #1077, opened 2024-11-17
Add BiConsumer to spring cloud stream plugin

LeovR:
This addresses #1071 but only adds the `BiConsumer`.

### DIFF
--- a/springwolf-examples/springwolf-cloud-stream-example/src/main/resources/application.properties
+++ b/springwolf-examples/springwolf-cloud-stream-example/src/main/resources/application.properties
@@ -10,7 +10,7 @@ spring.cloud.stream.default-binder=kafka
 spring.cloud.stream.binders.kafka.type=kafka
 spring.cloud.stream.binders.kafka.environment.spring.kafka.bootstrap-servers=${spring.kafka.bootstrap-servers}
 spring.cloud.stream.kafka.binder.autoCreateTopics=true
-spring.cloud.function.definition=process;consumerMethod;consumerClass;googlePubSubConsumerMethod
+spring.cloud.function.definition=process;consumerMethod;consumerClass;googlePubSubConsumerMethod;biConsumerMethod
 
 spring.cloud.stream.bindings.process-in-0.destination=example-topic
 spring.cloud.stream.bindings.process-in-0.group=springwolf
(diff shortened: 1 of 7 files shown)


LeovR:
I haven't yet used the `BiFunction` myself. Thus, I only added the `BiConsumer`, since I'm familiar with that.
I think I should be possible in the same manner.
I also tested the refactoring with the `FunctionCatalog` [here](https://github.com/springwolf/springwolf-core/commit/d8264d6ec639244182df903d0a62f6555e50a744).
But this is kind of a bigger refactoring and does not play well with the configured base-packages.

springwolf-examples/springwolf-cloud-stream-example/src/main/resources/application.properties
```diff
@@ -39,7 +42,11 @@ springwolf.docket.info.contact.url=https://github.com/springwolf/springwolf-core
 springwolf.docket.info.license.name=Apache License 2.0
 springwolf.docket.servers.kafka-server.protocol=kafka
 springwolf.docket.servers.kafka-server.host=${spring.cloud.stream.binders.kafka.environment.spring.kafka.bootstrap-servers}
-springwolf.use-fqn=false
```
timonback, line 42: Originally, we used cloud-stream to demo (and integration test) use-fqn=false,
but it could be the time to switch this.
I opened https://github.com/springwolf/springwolf-core/pull/1078 for a dedicated test to keep the test coverage

timonback:
The change looks good to me, even when headers are not supported yet, it is a step in the right direction.

Have a look at the publishing functionality, if there is no cloudstream way available, then lets leave it (+ useFqn) unchanged.

springwolf-examples/springwolf-cloud-stream-example/src/main/resources/application.properties
LeovR: I wasn't able to publish without this, because of `ClassNotFoundException`s

LeovR:
@timonback one could use the `StreamBridge` for publishing in the cloud stream context.
See e. g. [here](https://docs.spring.io/spring-cloud-stream/reference/spring-cloud-stream/producing-and-consuming-messages.html#sending-arbitrary-data-to-an-output-e-g-foreign-event-driven-sources)

timonback:
> I also tested the refactoring with the FunctionCatalog [here](https://github.com/springwolf/springwolf-core/commit/d8264d6ec639244182df903d0a62f6555e50a744).

I like the change, it seems to be cloud-stream way.
Feel free to go ahead - in a seperate PR.

(Is a dependency required? I am not able to autowire it in the cloudstream-example.)

timonback:
> @timonback one could use the `StreamBridge` for publishing in the cloud stream context. See e. g. [here](https://docs.spring.io/spring-cloud-stream/reference/spring-cloud-stream/producing-and-consuming-messages.html#sending-arbitrary-data-to-an-output-e-g-foreign-event-driven-sources)

Looks like an interesting starting point to browse the docs. Could be a future enhancement.

springwolf-examples/springwolf-cloud-stream-example/src/main/resources/application.properties
timonback: Fair. You can combine the cloud-stream and kafka plugin in your project for it.

I guess the project setup is not documented well enough.
Each example projects allows us to test/verify that its matching plugin works as expected.

Unfortunately, `core` has limited spring boot integration test capabilities - something we are working on currently. Therefore, the `example` projects serve as integration tests as well and are independent of each other.

PR #1087, opened 2024-11-20
Add BiFunction support to spring-cloud-stream-plugin

LeovR:


timonback (approved):
@LeovR When looking at the documentation [0], spring may bind 'anything' that matches - that can be headers, but also another topic (`-in-2`).
Is there a way to figure this out, for example via the FunctionCatalog?

> if (BiFunction.class.isAssignableFrom(type)) {
>            Type inputType = getTypeGenerics(element).get(0);
>.           // .get(1) could be used in the future
>            Type outputType = getTypeGenerics(element).get(2);

---

In any case, this is an improvement to the current state.
Thank you for this contribution.

[0] https://docs.spring.io/spring-cloud-stream/reference/kafka/kafka-streams-binder/programming-model.html#multiple-input-bindings (assuming this applies not only for KStream)

LeovR:
I think I have to dig deeper into the Kafka Streams Binder which behaves differently compared to the regular Kafka Binder.
At first look it seems that the Kafka Streams Binder does not use the FunctionCatalog like the regular Kafka Binder does.
The semantic seems to be different for the BiFunction types depending on the binder implementation.

timonback:
Fair enough, KStreams might be completely different anyway.

Since I do not use cloud stream in production, my knowledge is limited and I am very grateful for your input.

I'll merge the PR as it is good as it is right now. Feel free to open more PRs, and/or reopen the original issue.

PR #1273, opened 2025-04-13
draft: feat: support for azure service bus

Silvarr97:
feat: initial commit for Azure Service Bus plugin proposal

This commit introduces a non-runnable draft of an Azure Service Bus plugin integration.
The purpose of this code is to demonstrate the general direction I intend to take in order to support message publishing from the UI for testing purposes.

We currently use Azure Service Bus in our projects and found it limiting that we couldn't use the existing UI-based message publishing for testing. Since the Azure Service Bus Java SDK does not support JMS and communicates over AMQP 1.0, a dedicated integration is required.

I would like to contribute this feature to the project and would appreciate guidance and feedback from the maintainers. Specifically, I’d like to confirm if this is a suitable approach before proceeding with a functional implementation.


sam0r040:
Hi @Silvarr97, 
thanks for opening a draft PR. Currently we only find the producer configuration for Springwolf, but lack the scanners and bindings that we would expect from a plugin. Based on our conversation on Discord, it seems that `@AsyncPublisherAnnotation` is sufficient for you. Similarly the `@AsyncGenericOperationBinding` supports all bindings including ASB. 
We can link to the producer class from the website, so that others can copy it. Every new plugin requires more maintenance efforts for us a maintainers, so we need to balance maintenance work for us vs implementation effort for our users. In this case we do not see that adding a custom producer is a lot of overhead and can be done with Springwolf in your project already.

Issue #1425, opened 2025-09-19
SpringWolf-UI's resourceHandler breaks existing resourceHandlers


rfelgent:
**Describe the bug**
After adding SpringWolf-UI to one of my apps,  the "static resource caching" is broken. 
It seems to me the `SpringwolfUiResourceConfigurer` is causing this issue by adding its own `resourceHandlers()` 

```
   @Override
    public void addResourceHandlers(ResourceHandlerRegistry registry) {
        String springwolfBasePath = springwolfConfigProperties.getPath().getBase();

        log.debug("Serving Springwolf with base-path: {}", springwolfBasePath);

        registry.addResourceHandler(springwolfBasePath + "/**", webMvcProperties.getStaticPathPattern())
                .addResourceLocations(buildStaticLocation());
    }
```

And re-applying `webMvcProperties.getStaticPathPattern()` in SpringWolf' overwrites the *staticPathPattern* within `WebMvcAutoConfiguration`  and therefore turns off the static resource caching, e.g.

```
...
@Override
 public void addResourceHandlers(ResourceHandlerRegistry registry) {
      if (!this.resourceProperties.isAddMappings()) {
        logger.debug("Default resource handling disabled");
      } else {
        this.addResourceHandler(registry, this.mvcProperties.getWebjarsPathPattern(), "classpath:/META-INF/resources/webjars/");
        this.addResourceHandler(registry, this.mvcProperties.getStaticPathPattern(), (Consumer)((registration) -> {
          registration.addResourceLocations(this.resourceProperties.getStaticLocations());
          if (this.servletContext != null) {
            ServletContextResource resource = new ServletContextResource(this.servletContext, "/");
            registration.addResourceLocations(new Resource[]{resource});
          }

        }));
      }
    }
...
private void addResourceHandler(ResourceHandlerRegistry registry, String pattern, Consumer<ResourceHandlerRegistration> customizer) {
      if (!registry.hasMappingForPattern(pattern)) {
        ResourceHandlerRegistration registration = registry.addResourceHandler(new String[]{pattern});
        customizer.accept(registration);
        ////////////////////////////////////////////////////////////////////////////
        // this cache config is getting lost because of SpringWolf
        ////////////////////////////////////////////////////////////////////////////
        registration.setCachePeriod(this.getSeconds(this.resourceProperties.getCache().getPeriod()));
        registration.setCacheControl(this.resourceProperties.getCache().getCachecontrol().toHttpCacheControl());
[...]

PR #1435, opened 2025-09-26
bugfix(ui): only add ressource handler for custom springwolf paths

timonback:
The resourceHandler is additive.
SpringwolfUiResourceConfigurer allows to run springwolf-ui when the base-path is non-default


timonback:
relates to https://github.com/springwolf/springwolf-core/issues/1425

PR #1493, opened 2025-10-30
fix: update ReferenceUtil to replace # and * in  IDs

### LINKED ISSUE #1481: Support for # and *
Seems that # character cannot be escaped inside an AsyncAPI $ref path.
It's related with JSON Pointer (RFC 6901) and URI fragment (RFC 3986) specs

Still # and * are valid wildcards on ActiveMQ Artemis:
 
https://activemq.apache.org/components/artemis/documentation/latest/wildcard-syntax.html

Is there some workaround or something that lib can do to support it?

HamzehAdawi:
Fixes #1481 

Updated ReferenceUtil to replace #, and * with _
Added tests in ReferenceUtilTest.java to cover all three characters

### DIFF
--- a/springwolf-asyncapi/src/main/java/io/github/springwolf/asyncapi/v3/model/ReferenceUtil.java
+++ b/springwolf-asyncapi/src/main/java/io/github/springwolf/asyncapi/v3/model/ReferenceUtil.java
@@ -1,35 +1,32 @@
 // SPDX-License-Identifier: Apache-2.0
 package io.github.springwolf.asyncapi.v3.model;
 
-public class ReferenceUtil {
-    private static final String FORBIDDEN_ID_CHARACTER = "[#*/]";
+import java.util.Set;
+import java.util.stream.Collectors;
 
-    // Keeps valid characters, valid percent-encoded sequences,and changes forbidden to "_"
+public class ReferenceUtil {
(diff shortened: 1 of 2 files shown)


timonback:
Hi @HamzehAdawi 
Thank you for picking up this bug.

Great to see a test for this as well. I just triggered the ci.

The PR currently is very specific to the bug report.
Can you update it so that only the allowed characters from the rfc are allowed? That way, the solution fixes also potential follow up bugs with other illegal characters.

From the rfc:

> fragment    = *( pchar / "/" / "?" )

timonback:
Looks great, and fixes the issue in a generic way.

A documentation like to the rfc can be helpful, so that it is clear how the characters array was created, still, it can be found via the git blame.

Thank you for contributing and making Springwolf better.

HamzehAdawi:
Hey, @timonback so I realized there’s an edge case where a character may appear illegal but is actually valid when percent-encoded which I think was the main problem with the person's original issue regarding # and *.

Also, sorry could not quite get the allowed-characters approach to work (I think some tests expected illegal characters to be removed, but my version was replacing them with _ ) so I settled for this and it handles forbidding the literals #, * and / characters but permitting percent encoded.  

I hope this helps!

Issue #1291, opened 2025-05-12
[Enhancement request] support other schema formats, especially openapi schema.


tvahrst:
Since V3, AsyncApi supports different possible [schema formats](https://www.asyncapi.com/docs/migration/migrating-to-v3#schema-format-and-schemas) to describe the payload of messages (and headers). 

While `application/vnd.aai.asyncapi;version=3.0.0` is the default schema format, there are some other allowed formats for payload description:

[MultiSchemaObject](https://www.asyncapi.com/docs/reference/specification/v3.0.0#multiFormatSchemaObject)

SpringWolf currently delivers AsyncApis with data schemas only in default schema format (`application/vnd.aai.asyncapi;version=3.0.0`). My suggestion is to provide support for alternative schema formats, beginning with `application/vnd.oai.openapi;version=3.0.0` which is one format in the list of the 'recommended' schema formats an asyncapi implementation should support. 

**Motivation**
While `application/vnd.aai.asyncapi;version=3.0.0` and `application/vnd.oai.openapi;version=3.0.0` are quite similar, they differ in some details, especially the format of discriminator informations for inheritance. 

We do currently use [openapi generator](https://github.com/OpenAPITools/openapi-generator) to generate java model classes from openapi and we would like to use the same generator to generate model classes from asyncapi files (extracting only the components/schemas parts). Which is currently not possible because an asyncapi schema  with discriminator attributes can not be unmarshalled to a valid OpenAPI Model. 

**Technical details**
Under the hood, SpringWolf uses currently Swagger ModelResolver and ModelConverters to generate OpenApi/Swagger Schema Models, which are then transformed to SchemaObject Types (Springwolf internal *asyncapi* schema implementation). So there is allready an existing routine to generate `application/vnd.oai.openapi;version=3.0.0`.  Additionaly, SpringWolf has `MultiFormatSchema` class as Container for a schema-Reference and an schemaformat Field. Event the Enum `SchemaFormat` does exist (however openapi-schema still missing). 

A possible solution introduce a new environment property for the global schema format (defaults to asyncapi-schema) and optional allowing 'openapi-schema'.  On a first glance, these classes needs an update:

* SwaggerSchemaService
* SwaggerSchemaUtil (?)
* ExtractedSchemas 
  
We would like to provide a PR for this enhancement.

sam0r040:
Hi @tvahrst, 
sounds like a good enhancement for Springwolf. At the moment we only have two questions:
[...]

PR #1494, opened 2025-11-02
Support for openapi v3/v3.1 schema formats #1291

tvahrst:
I introduced a new Property `springwolf.docket.payload-schema-format` which is an enum and supports currently

* asyncapi_v3  (the default)
* openapi_v3
* openapi_v3_1

With the default-setting (asyncapi_v3), SpringWolf behaves  as before. 

Enabling openapi_v3 or _v3_1 leads to a AsyncApi where the **payload** schemas (not the header schemas!) are formatted with openapi v3/v3.1 format. This is possible and allowed because the AsyncApi Spec supports the definition of alternative schema formats.

The openapi v3/v3.1 schema formats are especially helpful when it comes to polymorphic schemas with custom discrimiantor mappings (which are not possible to describe with default asyncapi schema format).

### DIFF
--- a/springwolf-plugins/springwolf-cloud-stream-plugin/src/test/java/io/github/springwolf/plugins/cloudstream/asyncapi/scanners/channels/CloudStreamFunctionChannelsScannerIntegrationTest.java
+++ b/springwolf-plugins/springwolf-cloud-stream-plugin/src/test/java/io/github/springwolf/plugins/cloudstream/asyncapi/scanners/channels/CloudStreamFunctionChannelsScannerIntegrationTest.java
@@ -31,8 +31,8 @@ import io.github.springwolf.core.asyncapi.scanners.common.payload.internal.Paylo
 import io.github.springwolf.core.asyncapi.scanners.common.payload.internal.PayloadService;
 import io.github.springwolf.core.asyncapi.scanners.common.payload.internal.TypeExtractor;
 import io.github.springwolf.core.asyncapi.schemas.ModelConvertersProvider;
-import io.github.springwolf.core.asyncapi.schemas.SwaggerSchemaService;
 import io.github.springwolf.core.asyncapi.schemas.SwaggerSchemaMapper;
+import io.github.springwolf.core.asyncapi.schemas.SwaggerSchemaService;
 import io.github.springwolf.core.configuration.docket.DefaultAsyncApiDocketService;
 import io.github.springwolf.core.configuration.properties.SpringwolfConfigProperties;
 import io.github.springwolf.plugins.cloudstream.asyncapi.scanners.common.FunctionalChannelBeanBuilder;
--- a/springwolf-plugins/springwolf-jms-plugin/src/test/java/io/github/springwolf/plugins/jms/controller/SpringwolfJmsControllerIntegrationTest.java
+++ b/springwolf-plugins/springwolf-jms-plugin/src/test/java/io/github/springwolf/plugins/jms/controller/SpringwolfJmsControllerIntegrationTest.java
@@ -12,8 +12,8 @@ import io.github.springwolf.core.asyncapi.scanners.common.payload.internal.Paylo
 import io.github.springwolf.core.asyncapi.scanners.common.payload.internal.PayloadService;
 import io.github.springwolf.core.asyncapi.scanners.common.payload.internal.TypeExtractor;
 import io.github.springwolf.core.asyncapi.schemas.ModelConvertersProvider;
-import io.github.springwolf.core.asyncapi.schemas.SwaggerSchemaService;
 import io.github.springwolf.core.asyncapi.schemas.SwaggerSchemaMapper;
+import io.github.springwolf.core.asyncapi.schemas.SwaggerSchemaService;
 import io.github.springwolf.core.configuration.properties.SpringwolfConfigProperties;
 import io.github.springwolf.core.controller.PublishingPayloadCreator;
 import io.github.springwolf.core.controller.dtos.MessageDto;
--- a/springwolf-plugins/springwolf-kafka-plugin/src/test/java/io/github/springwolf/plugins/kafka/controller/SpringwolfKafkaControllerIntegrationTest.java
+++ b/springwolf-plugins/springwolf-kafka-plugin/src/test/java/io/github/springwolf/plugins/kafka/controller/SpringwolfKafkaControllerIntegrationTest.java
@@ -12,8 +12,8 @@ import io.github.springwolf.core.asyncapi.scanners.common.payload.internal.Paylo
 import io.github.springwolf.core.asyncapi.scanners.common.payload.internal.PayloadService;
 import io.github.springwolf.core.asyncapi.scanners.common.payload.internal.TypeExtractor;
 import io.github.springwolf.core.asyncapi.schemas.ModelConvertersProvider;
-import io.github.springwolf.core.asyncapi.schemas.SwaggerSchemaService;
 import io.github.springwolf.core.asyncapi.schemas.SwaggerSchemaMapper;
+import io.github.springwolf.core.asyncapi.schemas.SwaggerSchemaService;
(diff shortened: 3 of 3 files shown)


springwolf-core/src/test/java/io/github/springwolf/core/integrationtests/AsyncApiDocumentWithOpenApiV31SchemaFormatIntegrationTest.java
```diff
@@ -0,0 +1,437 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.core.env.ConfigurableEnvironment;
+import org.springframework.test.context.TestPropertySource;
+
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class AsyncApiDocumentWithOpenApiV31SchemaFormatIntegrationTest {
```
timonback: review: do we need all the tests on this level?
I wonder if one happy path test case is sufficient. Feels like a lot of test duplication

springwolf-core/src/main/java/io/github/springwolf/core/configuration/properties/PayloadSchemaFormat.java
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: Apache-2.0
+package io.github.springwolf.core.configuration.properties;
+
+import io.github.springwolf.asyncapi.v3.model.schema.SchemaFormat;
+
+/**
+ * Enumeration defining the supported payload schema formats, for use in SpringwolfConfigProperties.
+ */
+public enum PayloadSchemaFormat {
```
timonback: We are not sure if it a good idea to directly map the payload schema into the SchemaFormat enum, which has more types.
It couples the values of this enum to the other other and leaves less room for configuration changes.


Similarly, it feels odd in the switch expression to have an else branch, that can never be reached - at this moment in time, but is required technically for the compiler.

Suggestion:
ASYNCAPI_V3, OPENAPI_V3, OPENAPI_V3_1;

And the PayloadSchemaFormat is used in the code, including the switch expression

springwolf-examples/springwolf-jms-example/src/test/java/io/github/springwolf/examples/jms/ApiWithOpenApiV31SchemaFormatIntegrationTest.java
```diff
@@ -0,0 +1,44 @@
+// SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
timonback: One second thought, we believe that we can skip this test.

The examples are particular show cases of the main functionality - not all types of configurations possible.
Second (out of scope), we can benefit from integration tests in core with the asyncapi.json file checked in (currently tested manual in the two classes that you duplicated).

Also, the actual mapping is done by swaggers ModelConverter (a library that we do not control). Therefore, we see it not as much in our test scope.

timonback (commented):
@sam0r040 and I reviewed the PR. We enjoyed the PR very much. The PR is surprisingly sleek and it shows that you have understood and found the best place for integrating this feature.

We added some comments, suggestions. If there are questions, let us know. Some comments probably also reflect the Springwolf code style.
FYI: We committed & pushed a couple smaller changes (i.e. avoiding breaking API change)

springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/schemas/SwaggerSchemaUtil.java
```diff
@@ -25,9 +25,9 @@
 @RequiredArgsConstructor
 public class SwaggerSchemaUtil {
 
-    public Map<String, SchemaObject> mapSchemasMap(Map<String, Schema> schemaMap) {
+    public Map<String, ComponentSchema> mapSchemasMap(Map<String, Schema> schemaMap, SchemaFormat schemaFormat) {
```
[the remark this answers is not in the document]
tvahrst: I think, this is not possible, because 'mapSchemaAsMap' is not only used for payload schemas but also for header schemas which are always asyncapi-schemas. See
`SwaggerSchemaService#postProcessSimpleSchema` .

springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/schemas/SwaggerSchemaService.java
```diff
@@ -35,7 +36,9 @@
 
 @Slf4j
 public class SwaggerSchemaService {
-    private final ModelConverters converter = ModelConverters.getInstance();
+    private final ModelConverters converter_openapi30 = ModelConverters.getInstance(false);
+    private final ModelConverters converter_openapi31 = ModelConverters.getInstance(true);
+
```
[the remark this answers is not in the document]
tvahrst: Yes. I added a ModelConvertersProvider bean.

springwolf-core/src/test/java/io/github/springwolf/core/integrationtests/AsyncApiDocumentWithOpenApiV31SchemaFormatIntegrationTest.java
timonback: I see, still it feels like we are testing an external library.

Let's do the following: Include the updated tests as you suggest.
At the same time, https://github.com/springwolf/springwolf-core/pull/1509 is an alternative way to run these core integration tests (I agree that currently, there are no easy to use integration tests in core).

After this PR is merged, we / I re-visit this.

springwolf-core/src/main/java/io/github/springwolf/core/configuration/properties/PayloadSchemaFormat.java
timonback: Yeah, it is something in between.

From our point of view, the mapping from PayloadSchemaFormat to SchemaFormat happened to early. It is probably fine to have the getSchemaFormatForPayloadSchemaFormat method on the enum.

However, most of the code should use (and pass around) the configuration property (PayloadSchemaFormat). Which schema should be used depends on the configuration value (including the enum value) and not on the SchemaFormat enum (which might change, has more configuration options, etc).
Still, the SchemaFormat is used when using the schema format string, but it does not control program flow (switch)

springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/schemas/SwaggerSchemaService.java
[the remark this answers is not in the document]
timonback: Yes, that looks good.

Since the payloadSchemaFormat value cannot be changed, the constructor initializer can select (switch) the correct implementation so that `public ModelConverters getModelConverterForSchemaFormat(SchemaFormat schemaFormat) {` comes a simple `public ModelConverters getModelConverter() { return modelConverter;} `

springwolf-core/src/main/java/io/github/springwolf/core/configuration/properties/PayloadSchemaFormat.java
timonback: I pushed a commit to address/resolve it

timonback (commented):
I started looking at the output openapiv3 file it is looks quite similar. Besides that the json-schema extension is not really supported, the rest looks fine or related to the difference between openapi 3.0 vs 3.1

Testing:
-> remove runtimeOnly project(":springwolf-add-ons:springwolf-json-schema")
-> add @JsonPropertyOrder(alphabetic = true) to SchemaObject for easier comparison

---

One more question, when using the payloadSchemaFormat, why not also format the header schemas using the same setting?
Does your use-case only require the payloads in one format and the headers in the default?
I would have assumed that if I change it, all schema change.

PR #1510, opened 2025-11-09
feat: add format attribute to AsyncOperation.Headers.Header (#1506)

### LINKED ISSUE #1506: adding the option to specify datatypes and constraints for `@AsyncOperation.Headers.Header` in `io.github.springwolf.core.asyncapi.annotations`
I specify my AsyncAPI code first. When displaying my AsyncAPI spec with the web-ui (e.g. `http://localhost:8080/springwolf/asyncapi-ui.html`), the datatypes of the `payload` are correctly displayed. But for `header` fields, the displayed datatype is always `string` without any constraints, regardless what is defined in the actual code.

For example:

I have the following publisher:

```
import io.github.springwolf.bindings.kafka.annotations.KafkaAsyncOperationBinding;
import io.github.springwolf.core.asyncapi.annotations.AsyncOperation;
import io.github.springwolf.core.asyncapi.annotations.AsyncPublisher;

@Service
public class EventProducerService {

    @AsyncPublisher(operation = @AsyncOperation(
            channelName = "[...]",
            description = "[...]",
            servers = {"kafka-server"},
            payloadType = [...]Payload.class,
            headers = @AsyncOperation.Headers(
                    schemaName = "[...]Header",
                    values = {
                            @AsyncOperation.Headers.Header(name = "consultant-number", description = "Identifier for the consultant."),
                            @AsyncOperation.Headers.Header(name = "client-number", description = "Identifier for the client."))
                    }
            )
    ))
    @KafkaAsyncOperationBinding()
    public Mono<Void> publishEvent([...], final Integer consultantNumber, final Integer clientNumber) {

[...]
```
_specifics replaced with ` [...]`_

The actual datatypes for the header fields `consultant-number` and `client-number` are of type `Integer`, but  there is no way to specify that.

This is displayed as follows with the web-ui:

Image

Also, the generated JSON looks as follows:

```
  "components": {
    "schemas": {
      "[...]Header": {
        "title": "[...]Header",
        "type": "object",
        "properties": {
          "client-number": {
            "type": "string",
            "description": "Identifier for the client.",
            "enum": []
          },
          "consultant-number": {
            "type": "string",
            "description": "Identifier for the consultant.",
            "enum": []
          },
```

There seems to be no way to define the datatype as well as constraints with `@AsyncOperation.Headers.Header` or define the headers with a Java Class like the payload with `payloadType = [...]Payload.class`.

**Dependencies and versions used**
- [x] `springwolf-kafka` version `1.17.0`
- [x] `springwolf-ui` version `1.17.0`

**Motivation**
It should be possible to define specific datatypes and constraints for header fields to provide a correct specification of my AsyncAPI for consumers.

Haisia:
## Summary
Adds `format` field to `@AsyncOperation.Headers.Header` annotation to support AsyncAPI data type formats (e.g., int32, int64, date-time) per the AsyncAPI specification.

## Changes
- Added `format` attribute to `@AsyncOperation.Headers.Header` annotation with Javadoc
- Implemented `getFormat()` helper method in `AsyncAnnotationUtil`
- Updated `getAsyncHeaders()` to process and set format field
- Fixed `SwaggerSchemaUtil.mapSchemaObjectToSwagger()` - uncommented `setFormat()` call that was preventing format from appearing in output
- Added comprehensive tests for format handling
- Updated test hash values due to header structure change

## Implementation Note
During implementation, discovered that `SwaggerSchemaUtil.mapSchemaObjectToSwagger()` had `setFormat()` commented out, preventing format from appearing in the final AsyncAPI output. This has been fixed.

## Testing
Tested with `springwolf-kafka-example`:
- Headers with format display correctly in UI
- AsyncAPI JSON output includes format field
- Backward compatible (empty format = no format field)

**Example:**
```java
@AsyncOperation.Headers.Header(
    name = "user-id",
    description = "User identifier", 
    format = "int32"
)
```

**Output:**
```json
{
  "user-id": {
    "type": "string",
    "format": "int32",
    "description": "User identifier"
  }
}
```

Resolves #1506

### DIFF
--- a/springwolf-examples/springwolf-kafka-example/src/main/java/io/github/springwolf/examples/kafka/producers/AnotherProducer.java
+++ b/springwolf-examples/springwolf-kafka-example/src/main/java/io/github/springwolf/examples/kafka/producers/AnotherProducer.java
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 package io.github.springwolf.examples.kafka.producers;
 
+import io.github.springwolf.bindings.kafka.annotations.KafkaAsyncOperationBinding;
 import io.github.springwolf.core.asyncapi.annotations.AsyncOperation;
 import io.github.springwolf.core.asyncapi.annotations.AsyncPublisher;
 import io.github.springwolf.examples.kafka.configuration.KafkaConfiguration;
@@ -33,6 +34,7 @@ public class AnotherProducer {
                                                         description = "Event identifier",
                                                         format = "uuid")
                                             })))
+    @KafkaAsyncOperationBinding
     public void sendMessage(AnotherPayloadDto msg) {
         kafkaTemplate.send(KafkaConfiguration.PRODUCER_TOPIC, msg);
     }
--- a/springwolf-examples/springwolf-kafka-example/src/test/resources/asyncapi.json
+++ b/springwolf-examples/springwolf-kafka-example/src/test/resources/asyncapi.json
@@ -1953,7 +1953,11 @@
         },
         "name": "io.github.springwolf.examples.kafka.dtos.AnotherPayloadDto",
         "title": "AnotherPayloadDto",
-        "bindings": { }
+        "bindings": {
+          "kafka": {
+            "bindingVersion": "0.5.0"
+          }
+        }
       },
       "io.github.springwolf.examples.kafka.dtos.ExamplePayloadDto": {
         "headers": {
@@ -2174,7 +2178,11 @@
       },
       "title": "another-topic_send",
       "description": "Auto-generated description",
-      "bindings": { },
+      "bindings": {
+        "kafka": {
+          "bindingVersion": "0.5.0"
+        }
+      },
       "messages": [
         {
           "$ref": "#/channels/another-topic/messages/io.github.springwolf.examples.kafka.dtos.AnotherPayloadDto"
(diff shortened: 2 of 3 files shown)


springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/scanners/common/annotation/AsyncAnnotationUtil.java
```diff
@@ -100,6 +106,17 @@ private static String getDescription(
                 .orElse(null);
     }
 
+    private static String getFormat(
+            List<AsyncOperation.Headers.Header> value, StringValueResolver stringValueResolver) {
+        return value.stream()
+                .map(AsyncOperation.Headers.Header::format)
+                .filter(StringUtils::hasText)
+                .map(stringValueResolver::resolveStringValue)
```
timonback: This differs from the description case above. Is it possible to include the has text here as well to avoid the if block?

Or have you found an issue so that the description (and other fields) should be changed?

timonback (commented):
Looks great, Thank you @Haisia 

We use our Kafka example also for show casing and indirectly as full integration test.
You mentioned that you used it for verification as well. Let's update the Kafka example (one header format is sufficient) to demonstrate this feature.

springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/scanners/common/annotation/AsyncAnnotationUtil.java
Haisia: Thank you for the feedback!

You're right - I should align with the `getDescription()` pattern for consistency.

I initially used `filter → resolve` because the test mock adds "Resolved" to empty strings, but that's actually a test issue, not a code issue.
(I thought it was strange that the result "Resolved" is returned even when the value is null or empty.)

Issue #635, opened 2024-03-04
`use-fqn` overrides SpringDoc configuration


ctasada:
**Describe the bug**
In an application that uses both `SpringDoc` for OpenApi and `Spring-Wolf` for AsyncAPI, the usage of `spring-wolf.use-fqn` overrides the one used by `springdoc.use-fqn`

I just realized after upgrading an application to spring-wolf:core 1.0.0 since the `use-fqn` is now defaulting to `true` the issue may be older.

**Code example**
Having a SpringBoot application with both dependencies, and the application.yml configured like
```
springdoc:
    use-fqn: true
    api-docs:
        enabled: true
        
springwolf:
    enabled: true
    use-fqn: false
```

I would expect the `springdoc` generated OpenAPI uses the FQN, but it's not the case. The value used for `spring-wolf.use-fqn` is alway taking precedence.

**Stack trace and error logs**
None

timonback:
Thank you for the report and I agree with your expectation.

As part of the 1.0 release, I also slightly updated the code for handling the fqn setting.
Now, Springwolf always resets the `swagger-core` setting back to its original value: https://github.com/springwolf/springwolf-core/blob/a3b4743b825efa79added6d186ea4e581dce0f2b/springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/components/DefaultComponentsService.java#L175

We contributed the getter to `swagger-core` a while back in https://github.com/swagger-api/swagger-core/blame/3fcc473c33c56f306902853a451020926ad691ea/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/TypeNameResolver.java#L25
Unfortunately, it is a static field. So we set and reset it globally.

---

I wonder whether we are running into concurrency issues.

Is this also reproducible when the Springwolf `init-mode` is `background`?

sam0r040:
Hi @ctasada I had another look at it and stumbled on the configuration property `springdoc.pre-loading-enabled`. If it is set to true then then springdoc will create the spec asynchronously although they are only setting the value at the constructor level: https://github.com/springdoc/springdoc-openapi/blob/0fdfabadf9a51895ada8c432aa521715e971b149/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/service/OpenAPIService.java#L210

If thats the case, there might be issues with multithreading.

We can have a deeper look if you could provide an example that reproduces the issue.

ctasada:
I tried the different approaches and the issue persists.

[...]

PR #1709, opened 2026-04-19
fix: make usefqn threadsafe

tvahrst:
Unfortunately, using the static 'std' instance of Swagger's `TypeNameResolver` is not thread-safe. This can lead to unwanted race conditions and side effects (see Issue #635).

To fix this, code sections using `TypeNameResolver` must utilize a dedicated, properly configured instance (with usefqn set correctly). This requires the ability to create a custom instance of `TypeNameResolver`. Since the constructor is protected, a simple subclass `SpringWolfTypeNameResolver` was introduced to make the constructor public.

A custom `TypeNameResolver` must be implemented in the following locations:

* `SwaggerSchemaService`: Two `TypeNameResolver` instances are required—one to generate simple names and a second instance based on the 'useFqn' property.
* Swagger `ModelConverters`: Internally, this uses a `ModelResolver`,  which in turn relies on the global, static instance of `TypeNameResolver`. To correct this, the internal `ModelResolver` instance must be replaced with a properly initialized one during the creation of `ModelConverters`. See `ModelConvertersProvider`.


timonback:
I took a look, there is something wrong in combination with Protobuf or how the ObjectMapper is used to generated the examples.
I looks better now or it seems there was a bug before for the openapi31 format output. But for some reason, the ConsumerRecordMetadata schema has no type and therefore the example is empty / throws an exception.

@tvahrst Can you confirm, that the updated schema with openapi31 matches your expectation?
Any thoughts on the  ConsumerRecordMetadata?

tvahrst:
Hi @timonback , the Spring class `ConsumerRecordMetaData`, which is used as a `@Header` in the Kafka tests, doesn't conform to Java Bean conventions, so Swagger `ModelConverters` can't generate a correct schema from it. While `ModelConverters` 3.0 at least defined `type=object`, `ModelConverters` 3.1 generates an empty schema with `type=null`, which leads to the current result.

I've provided a separate, small ModelConverter class for the Kafka test suite that explicitly reacts to `ConsumerRecordMetaData` and provides a simple, empty object-schema. With this, the result of the `PayloadSchemaFormatIntegrationTest` for OpenAPI 3.1 looks quite similar to OpenAPI 3.0. 

While I was at it, I also made a small adjustment to the error output during example schema generation to display the name of the schema in case of exceptions.

I was wondering why the test for OpenAPI 3.1 changed so much compared to the master version due to this pull request. The reason was this: On the master version, a ModelConverter bean, specifically a ModelResolver, is created in the ProtobufStandAloneConfiguration.

Firstly, this is internally set to OpenAPI 3.0. Secondly, this is an 'additionalModelConverter', which is collected in the `ModelConvertersProvider` and assigned to both internal ModelConverters instances (3.0 and 3.1). Therefore, the 3.1 ModelConverter contained *two* ModelResolver instances and was inadvertently converted into a 3.0 ModelConverters bean.

timonback:
Good catch, so adding the ConsumerMetaData in general could be seen as wrong and could be omitted (in a future adjustment).

Adding more debug output is good. Ill have to check that that cannot raise a nullpointer exception.

Thank you for the explanation. Just to confirm, there was a bug before and it is corrected now? Or do you see another issue with the modelresolvers for protobuf?

tvahrst:
Yes, the new `ProtobufConfiguration` without providing a `ModelResolvers` bean looks good for me.

PR #1835, opened 2026-08-23
fix: handle null schemaType in assignType for composed schemas

### LINKED ISSUE #1834: NPE in SwaggerSchemaMapper.assignType when a property schema has no type (composed schema)
**Describe the bug**

`SwaggerSchemaMapper.assignType` calls `Set.of(schemaType)` when the swagger `Schema` has a null `type`. `Set.of` rejects null, so any property whose schema is a composed schema (`allOf` / `oneOf` / `anyOf`, which legitimately carry no `type`) aborts channel and operation scanning with an NPE.

The failure is caught and logged as a WARN, so the published AsyncAPI document silently loses every channel the scanner had not reached yet. In our service 9 of 15 channels disappeared, and the only signal was four WARN lines at startup — the message body is literally `null`, so there was nothing to identify the offending payload with.

`springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/schemas/SwaggerSchemaMapper.java` (unchanged on `main` at the time of writing):

```java
203  private static void assignType(Schema swaggerSchema, SchemaObject.SchemaObjectBuilder builder, boolean isNullable) {
204      Set<String> schemaTypes = swaggerSchema.getTypes();
205      Set<String> types = schemaTypes == null ? new HashSet<>() : new HashSet<>(schemaTypes);
206
207      String schemaType = swaggerSchema.getType();
208      if (!types.contains(schemaType)) {
209          // required while swagger v2 does not populate types
210          builder.type(Set.of(schemaType));   // NPE when schemaType == null
211          return;
212      }
```

Lines 205 and 208 already tolerate nulls (`HashSet.contains(null)` returns `false`), so a type-less schema is routed straight into the one call that cannot accept null.

**How we ran into it**

springdoc-openapi 3.1.0 added `KotlinNullablePropertyCustomizer`, which implements `io.swagger.v3.core.converter.ModelConverter` and is registered as a bean. `ModelConvertersProvider:40` adds every `ModelConverter` bean in the Spring context to springwolf's converters, so it gets applied to Kafka payload schemas too.

For a nullable Kotlin property of object type it replaces the property schema with a composed wrapper (`allOf: [$ref]` + `nullable` under OAS 3.0, `oneOf: [$ref, {type: "null"}]` under 3.1). That wrapper has neither `type` nor `$ref`, which is valid OpenAPI, and `assignType` then throws.

springdoc is just how we hit it — any converter emitting a type-less composed schema would do the same.

**Suggested fix**

```java
builder.type(schemaType == null ? Set.of() : Set.of(schemaType));
```

Separately, it would help a lot if `DefaultChannelsService` / `DefaultOperationsService` logged the schema or class being scanned when they catch an exception. The current WARN carries only `null`, which meant tracking this down required decompiling to find out which payload failed.

**Dependencies and versions used**
- [x] `springwolf-core` version `2.6.0` (code path unchanged on `main`)
- [x] `springwolf-kafka` version `2.6.0`
- [x] `springdoc-openapi-starter-webflux-api` version `3.1.0`
- [x] Spring Boot `4.1.0`, Kotlin `2.3.10`, JVM 25

**Code example**

Payload (any nullable, object-typed property will do):

```kotlin
data class AuthorisationPayload(
    val serviceReference: String,
    val merchant: Merchant?,   // nullable + object type -> composed schema with no type
) {
    data class Merchant(val settingsId: String)
}
```

Reproducer without booting Spring — this throws:

```kotlin
val properties = SpringwolfConfigProperties()
val provider = ModelConvertersProvider(
    properties,
    listOf(KotlinNullablePropertyCustomizer(ObjectMapperProvider(SpringDocConfigProperties()))),
)
val service = SwaggerSchemaService(properties, emptyList(), SwaggerSchemaMapper(properties), provider)

service.resolveSchema(AuthorisationPayload::class.java, "application/json")  // NullPointerException
```

In a running application the same thing happens during startup scanning, with the customizer picked up automatically as a bean. Removing that single bean (`springdoc.model-converters.kotlin-nullable-property-customizer.enabled: false`) restores all 15 channels, which is our current workaround.

**Stack trace and error logs**

```text
WARN i.g.s.c.a.c.DefaultChannelsService   : An error was encountered during channel scanning with
                                            ChannelsInClassScannerAdapter@2ada96fd: null
WARN i.g.s.c.a.o.DefaultOperationsService : An error was encountered during operation scanning with
                                            OperationsInClassScannerAdapter@41820105: null

java.lang.NullPointerException
    at java.base/java.util.Objects.requireNonNull(Objects.java:220)
    at java.base/java.util.ImmutableCollections$Set12.<init>(ImmutableCollections.java:1009)
    at java.base/java.util.Set.of(Set.java:470)
    at io.github.springwolf.core.asyncapi.schemas.SwaggerSchemaMapper.assignType(SwaggerSchemaMapper.java:210)
    at io.github.springwolf.core.asyncapi.schemas.SwaggerSchemaMapper.mapSwaggerSchemaToAsyncApiSchema(SwaggerSchemaMapper.java:100)
    at io.github.springwolf.core.asyncapi.schemas.SwaggerSchemaMapper.mapSchema(SwaggerSchemaMapper.java:72)
    at io.github.springwolf.core.asyncapi.schemas.SwaggerSchemaMapper.mapSchemaOrRef(SwaggerSchemaMapper.java:51)
    at io.github.springwolf.core.asyncapi.schemas.SwaggerSchemaMapper.lambda$mapSwaggerSchemaToAsyncApiSchema$1(SwaggerSchemaMapper.java:105)
    at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:214)
    at java.base/java.util.Iterator.forEachRemaining(Iterator.java:133)
    ...
```

itsmib:
## Problem
`SwaggerSchemaMapper.assignType` calls `Set.of(schemaType)` when the 
swagger Schema has a null type. `Set.of` rejects null, causing an NPE 
for any property whose schema is a composed schema (allOf/oneOf/anyOf) 
which legitimately carries no type.

This silently drops channels/operations from the published AsyncAPI 
document, the only signal being a WARN log with `null` as the message.

## Changes
- Added null check in `assignType` before calling `Set.of(schemaType)`
- Improved warning log in `DefaultChannelsService` and 
  `DefaultOperationsService` to show exception class name when 
  `getMessage()` returns null

## Test
Added `mapNullTypeForComposedSchema` test in `SwaggerSchemaMapperTest` 
to verify no NPE is thrown for schemas with null type.

Fixes #1834

### DIFF
--- a/springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/channels/DefaultChannelsService.java
+++ b/springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/channels/DefaultChannelsService.java
@@ -34,7 +34,11 @@ public class DefaultChannelsService implements ChannelsService {
                 Map<String, ChannelObject> channels = scanner.scan();
                 foundChannelItems.addAll(channels.values());
             } catch (Exception e) {
-                log.warn("An error was encountered during channel scanning with {}: {}", scanner, e.getMessage(), e);
+                log.warn(
+                        "An error was encountered during channel scanning with {}: {}",
+                        scanner,
+                        e.getMessage() != null ? e.getMessage() : e.getClass().getSimpleName(),
+                        e);
             }
         }
         return ChannelMerger.mergeChannels(foundChannelItems);
--- a/springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/channels/DefaultChannelsService.java
+++ b/springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/channels/DefaultChannelsService.java
@@ -34,11 +34,7 @@ public class DefaultChannelsService implements ChannelsService {
                 Map<String, ChannelObject> channels = scanner.scan();
                 foundChannelItems.addAll(channels.values());
             } catch (Exception e) {
-                log.warn(
-                        "An error was encountered during channel scanning with {}: {}",
-                        scanner,
-                        e.getMessage() != null ? e.getMessage() : e.getClass().getSimpleName(),
-                        e);
+                log.warn("An error was encountered during channel scanning with {}: {}", scanner, e.getMessage(), e);
             }
         }
         return ChannelMerger.mergeChannels(foundChannelItems);
--- a/springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/operations/DefaultOperationsService.java
+++ b/springwolf-core/src/main/java/io/github/springwolf/core/asyncapi/operations/DefaultOperationsService.java
@@ -34,11 +34,7 @@ public class DefaultOperationsService implements OperationsService {
                 Map<String, Operation> channels = scanner.scan();
                 foundOperations.addAll(channels.values());
             } catch (Exception e) {
-                log.warn(
-                        "An error was encountered during operation scanning with {}: {}",
-                        scanner,
-                        e.getMessage() != null ? e.getMessage() : e.getClass().getSimpleName(),
-                        e);
+                log.warn("An error was encountered during operation scanning with {}: {}", scanner, e.getMessage(), e);
             }
         }
         return OperationMerger.mergeOperations(foundOperations);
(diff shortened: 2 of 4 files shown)


timonback:
Hi @itsmib,
thank you for the fix and the included test to ensure that it is covered long term.

I am not sure about the benefit of logging the name of the exception, in case the message of the exception is null. The exception including stacktrace is logged anyway and part of the log.
The original issue is the missing knowledge, which method/payload triggered the issue. Something that is partly covered by the `DEBUG` level logging, but can be improved upon.

For this PR, can you revert the logging changes?
The fix and test look great!

itsmib:
Hey @timonback, Thanks for the review. Logging changes are reverted now.

